PR #49, opened 2015-03-12
Replace deprecated des_ and fix LibreSSL build

Sp1l:
uses deprecated des_ methods and types that were marked deprecated by
OpenSSL 0.9.7 and will be removed in OpenSSL 1.1.0 . This patch replaces
the des_ methods and types with their new DES_ counterparts. This
enables building with LibreSSL as OpenSSL library.

Check https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/crypto/des/des_old.h for the compatibility macros that have been removed.


vanhauser-thc:
that is useful! thanks!

PR #65, opened 2015-06-06
adding make location configurations to ease proper packaging

anthraxx:
this patch introduces DATADIR, MANDIR and DESTDIR to it is possible
to configure the csv data location, the man location and also a
packaging DESTDIR location which is needed for packaging in a subdir.
PREFIX stays like it was so it is properly used in dpl4hydra.sh.

The old locations and behavior should still be the same if you don't override any of the new variables, so this patch should not change any of the "old" default locations but adds the possibility to distribute to any arbitrary location (dpl4hydra.sh respects those by introducing another sed command for LOCATION).

Just as an example, I use this patch to package hydra for ArchLinux via:

    ./configure --prefix=/usr
    make
    make DESTDIR="${pkgdir}" MANDIR="/share/man/man1" DATADIR="/share/${pkgname}" install

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,7 +4,10 @@
 OPTS=-I. -O3
 # -Wall -g -pedantic
 LIBS=-lm
-DIR=/bin
+BINDIR = /bin
+MANDIR ?= /man/man1
+DATADIR ?= /etc
+DESTDIR ?=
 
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
Thanks!

Issue #68, opened 2015-06-29
Add gzip support for login and passwords files.


renatoalencar:
For big wordlists, it could be compressed in gzip format. So, I propose add gzip support for used files. By the way, I am working on it.

vanhauser-thc:
patch has been submitted and applied

PR #69, opened 2015-07-01
Add gzip support as proposed in issue #68

renatoalencar:
Just let gzip functions of zlib read files. When in gzip format these functions decompress files, if not in gzip format, just read as decompressed stream.


vanhauser-thc:
Hi,
sorry, I am ill at the moment ...
If I understand the patch correctly, instead of normal files it only supports gzip'ed files now?
Or does the gzopen detect if its a normal file or gziped one?
The libz manual is not clear about that.

vanhauser-thc:
ah, sorry, now I see your description text above :)
as I said, I am ill at the moment with fever :)
sorry, to be sure, with the patch it does not matter if the files are gziped or not and it will work?

renatoalencar:
"gzopen can be used to read a file wich is not in gzip format; in this case gzread will directly read from the file without decompression. When reading, this will be detected automatically by looking for the magic two-byte gzip header."

renatoalencar:
This is what i found about when i was reading the documentation, it can be found in zlib.h. gzopen reads the first bytes in stream and detect if it's 1F8B. Thanks for feedbacks :)

vanhauser-thc:
OK, let me rephrase that - have you actually tried that both work, gzip'ed and normal text file?
:)

renatoalencar:
Yes, i used with a wordlist that comes with Kali Linux in /usr/share/wordlists/rockyou.txt.gz, gzip'ed and uncompressed, it worked right at the moment.

vanhauser-thc:
thanks!

PR #70, opened 2015-07-09
Allow compilation with older openssl

allfro:
Uses `TLSv1_client_method` if `TLSv1_2_client_method` is not defined as a symbol in OpenSSL.


vanhauser-thc:
useful, thanks!

PR #81, opened 2015-11-26
Fix missing openssl library on android 6.0

tux-mind:
from android 6.0 google throw away the openssl library to use a lighter ssl library.

with this change openssl will be compiled statically into hydra, making it independent from underlay ssl implementation.


vanhauser-thc:
thanks!

PR #82, opened 2015-11-29
use static version of the ssl library

tux-mind:
sorry for coming out later with these commits, I installed an android 6.0 emulator and tested out the whole stuff.

it should be `libssl_static`, not `libssl`, and it must be before `libcrypto_static`.

hydra define the function `ssl_cert_free`, that is also defined in openssl library.
solution to this problem is to define that function as file-scoped ( static ).

this will also fix hydra for anyone that want to build it statically with openssl library built-in.

thank you for your patience @vanhauser-thc :blush:

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -1,7 +1,7 @@
 LOCAL_PATH:= $(call my-dir)
 include $(CLEAR_VARS)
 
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
this is a -dev version of hydra, so it is OK to have incremental fixes :)

Issue #80, opened 2015-11-04
Custom SSL prefix


DomT4:
Something of a feature request - The ability to pass a custom prefix to the SSL libraries/headers.

As you may well be aware Apple removed the OpenSSL headers from OS X 10.11 and Xcode 7 (Which impacts OS X 10.10 as well now) and now `hydra` refuses to compile on those platforms with SSL support *unless* you have an SSL installation in another default-searched path as defined by `configure`. 

It's not always practical to put things in specific search paths so was wondering if you'd consider adding support for a custom SSL prefix?

I tried to throw something together in line with your current usage but my patience with `configure` scripts and autotools in general is not the highest and it doesn't quite work.

```diff
diff --git a/configure b/configure
index 8980e24..ef6e0ab 100755
--- a/configure
+++ b/configure
@@ -8,6 +8,8 @@ if [ "$1" = "-h" -o "$1" = "--help" ]; then
   echo "  --prefix=path              path to install hydra and its datafiles to"
   echo "  --with-oracle=prefix       prefix for oracle include dir"
   echo "  --with-oracle-lib=prefix   prefix for oracle lib dir"
+  echo "  --with-ssl=prefix          prefix for SSL headers"
+  echo "  --with-ssl-lib=prefix      prefix for SSL libraries"
   echo "  --disable-xhydra           disable compilation of hydra GUI"
   echo "  --nostrip                  do not per default strip binaries before install"
   echo "  --debug                    show debug output to trace errors"
@@ -25,6 +27,8 @@ WORACLE_PATH=""
 WORACLE_LIB_PATH=""
 SSL_PATH=""
 SSL_IPATH=""
+WSSL_PATH=""
+WSSL_LIB_PATH=""
 CURSES_PATH=""
 CURSES_IPATH=""
 CRYPTO_PATH=""
@@ -145,10 +149,14 @@ if [ "X" != "X$DEBUG" ]; then
    echo DEBUG: SSL_INC=$INCDIRS `ls -d /*ssl/include /opt/*ssl/include /usr/*ssl/include /usr/local/*ssl/include 2> /dev/null`
 fi
 
-for i in $LIBDIRS \
-/*ssl /usr/*ssl /opt/*ssl /usr/local/*ssl /opt/local/*ssl \
-/*ssl/lib /usr/*ssl/lib /opt/*ssl/lib /usr/local/*ssl/lib /opt/local/*ssl/lib
-do
[...]

PR #84, opened 2015-12-28
Fix issue with hydra not compiling since adding custom ssl prefix

dirkjanm:
As mentioned by users in #80, since adding the prefix the compilation is broken unless the SSL prefix is used.


vanhauser-thc:
thank you!

PR #85, opened 2016-01-01
If prefix set then high chance of libs being there

x0x7:
If they are compiling to target an alternative directory there are likely dependencies to the same target.  Test for PREFIX/lib and PREFIX/include to be sure.


vanhauser-thc:
good idea, merged.

PR #93, opened 2016-02-06
Sleeps are now interuptible and posix

x0x7:
usleep is now depricated and is uninteruptable.  nanosleep is considered the suggested replacement.  Unfortunately it isn't available in Windows but by writing seperate code we can add Window's interuptable SleepEx function.

nanosleep also can take seconds directly so there is an addition function written for that.


vanhauser-thc:
I do not see that still using usleep would be an issue, but as you made the effort of course I accept the patch :)
thanks!

PR #109, opened 2016-04-09
Fix several undefined behaviors

Diadlo:
By standard C 11 if sprintf has same source and destination it is UB. Fixed by adding intermediate buffer


vanhauser-thc:
thanks

Issue #116, opened 2016-05-11
http-post-form cuts off last character of redirect location


darrencocco:
Analyzed the output in debug mode and noticed that it was cutting down /login/index.php to /login/index.ph and /login/index.php?sessionid=2 to /login/index.php?sessionid=

Command run: 
./hydra example.com https-form-post "/login/index.php:username=^USER^&password=^PASS^:S=session has timed out" -l admin -P ~/passwords -t 64 -f -v -d

This is against the latest master commit as of the date of writing

vanhauser-thc:
duplicate and fixed

PR #117, opened 2016-05-11
Fix for #116

darrencocco:
Off by one error found when processing Location: in redirects. on lines 792-793

Pointer handling mistake found when updating cookies after redirect.
When the process_cookies function gets called it expects a pointer to a pointer to a cookie_node, instead it gets a pointer to a cookie_node, derefs the cookie_node struct and lands in the middle of invalid memory.


vanhauser-thc:
thanks!

darrencocco:
Happy to contribute.

PR #121, opened 2016-05-26
Add 'http-post', remove HTTP 403 as success criteria

hdm:
This is a bit of a hack and I could use some feedback rather than a merge here. I was testing a device that needed HTTP POST (no data) for the 401 response and sent back a 403 on error. This feels like it should be a bigger patch to make all of this configurable, but it works for me (tm) at the moment.

Happy to keep this as a side thing, but also up for a rework to make this configurable if there is any desire. 

Thoughts?

Thanks!

### DIFF
--- a/hydra-http.c
+++ b/hydra-http.c
@@ -192,7 +192,9 @@ int start_http(int s, char *ip, int port, unsigned char options, char *miscptr, 
   ptr = ((char *) index(http_buf, ' '));
   if (ptr != NULL)
     ptr++;
-  if (ptr != NULL && (*ptr == '2' || *ptr == '3' || strncmp(ptr, "403", 3) == 0 || strncmp(ptr, "404", 3) == 0)) {
+// Make this optional
+//  if (ptr != NULL && (*ptr == '2' || *ptr == '3' || strncmp(ptr, "403", 3) == 0 || strncmp(ptr, "404", 3) == 0)) {
+  if (ptr != NULL && (*ptr == '2' || *ptr == '3' || strncmp(ptr, "404", 3) == 0)) {
     hydra_report_found_host(port, ip, "www", fp);
     hydra_completed_pair_found();
     if (http_buf != NULL) {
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
Hey HDM, long time!

sorry for the late reaction, there is a lot on my plate at the moment.

so here is my feedback:

I see more and more weird wifi/dsl router companies doing stuff that is not RFC compliant because their developers are clueless.

One example is yours, another is an implementation which requries cookies(!) with HTTP basic auth.

I still do not know how to handle this. It is specific to a device or brand, and otherwise pretty useless.

I a thinking about something like http-custom-BRANDNAME, which then runs the normal "service_http_get" (or _post in your case) which handles the exceptions with if()'s.

Your patch adds the first basic requirement (although one more wrapper for http_custom_BRAND() would be required to point to http_post and set a global variable), your patch does additionally change the success criteria. this change however is specific to your device, and should only be enable if the custom global var is set.

I think that would be the best approach.
your thoughts?

hdm:
@vanhauser-thc 

Thank you for the feedback! I have been mulling on this today and I worry that even "http-custom-BRANDNAME" may not be specific enough (tons of products from the same vendor, multiple versions of a product over the years, etc) and would be a pain to maintain. What about exposing a bunch of configuration options for http (method, success-criteria, success-match, etc) and then building up a text file full of definitions for specific products? Someone would specify whatever crazy args are needed for the product they are testing that day, or just the name of the profile in the text file, which would load the parameters needed for that specific product-model-version.

vanhauser-thc:
well you could also do http-custom-VENDOR-PRODUCT.

what you are describing would be nice but also very, very complicated.

It first would not fit in hydra but would be a different/new tool and second configuring this alone would require a lot of HTTP and security understanding and time that in my experience very few people would actually use it.
While coding all this being very time consuming.
So everyone who could and would use it, would rather code quickly a perl/ruby/python script to do that special weird stuff.

vanhauser-thc:
btw - I just added http-post (but without the 403 change).
and to support SSL SNI I had to change nearly all C files and central connection functions.

hdm:
Thanks for the heads up! Will close this out for now, appreciate the feedback!

PR #122, opened 2016-05-27
Update hydra.c

lianchi:
In line 3907, added an `if` statement to prevent the overflow of long unsigned int value "-1".


renatoalencar:
Hey man. You even make any logging message, just exit. It could be a little bit boring to the user if the application just stop without warn anything.

vanhauser-thc:
Yes, I can not use the patch like this. It is useful though, so I will make a few changes. Still I have no clue when/how this can happen (and that is actually the bug that needs to be fixed)

vanhauser-thc:
I just commited a patch that exits with an error message that tells the user to post the commandline triggering the bug.
thanks!

PR #136, opened 2016-06-23
Patch for xHydra SNMP

PetarKaleychev:
The patch fixes wrong behavior in xHydra, for SNMP for versions 1 and 2.
It also adds version 3 and options as MD5, SHA, AES, READ, …


vanhauser-thc:
thank you!

PR #161, opened 2016-10-09
add LibreSSL support

sizeofvoid:
- Tested with LibreSSL 2.5.0 on OpenBSD current amd64


vanhauser-thc:
thanks a lot! that was on my wish list :)

PR #162, opened 2016-10-10
Fix compiler warning.

crondaemon:
When used with other statements, #ifdef should be split in tokens

Warning:
hydra-mod.c:441:19: warning: extra tokens at end of #ifdef directive


crondaemon:
Ops, sorry, other changes went into the commit. Pushing a new change.

PR #164, opened 2016-10-11
Bug fix and module improvement

PetarKaleychev:
The patch fixes a bug in the cases when the number of the tasks is smaller than the number of the passwords. Also includes improvement to support the features of the protocol.


vanhauser-thc:
thanks!

PR #170, opened 2016-11-07
Fix telnet complete condition

deniswhite8:
Fix false success for case like
`% Bad password`
`Invalid login/password`


vanhauser-thc:
thanks

PR #175, opened 2016-12-29
SunOS compile support

CaledoniaProject:
changes summary

```
remove GCCSEC options to compile on sunos
add -lrt to support nanosleep()
add definition of strndup()
```


vanhauser-thc:
thanks!

PR #185, opened 2017-01-31
Remove rindex() and always use strrchr()

fornwall:
The rindex() function was deprecated a long time ago and using strrchr() nowadays is safe, so maintaining a rindex() path is just unnecessary complexity.

Also, this fixes a build issue when cross compiling for Android using a standalone toolchain, as the probing in configure for rindex() uses the native gcc instead of the cross compiler.

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -2,7 +2,7 @@ LOCAL_PATH:= $(call my-dir)
 include $(CLEAR_VARS)
 
 LOCAL_CFLAGS:= -O3 -DLIBOPENSSL -DLIBIDN -DHAVE_PR29_H -DHAVE_PCRE \
-               -DLIBNCP -DLIBPOSTGRES -DLIBSVN -DLIBSSH -DNO_RINDEX \
+               -DLIBNCP -DLIBPOSTGRES -DLIBSVN -DLIBSSH \
                -DHAVE_MATH_H -DOPENSSL_NO_DEPRECATED -DNO_RSA_LEGACY \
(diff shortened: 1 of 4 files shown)


vanhauser-thc:
I don't really see the necessity to remove rindex, as I highly doubt it will be removed in the next 10 years, deprecated or not.

But I made the following changes:
1. now configure checks if strrchr is present and use it over rindex
2. ./configure now uses the CC environment variable if present

CC=gcc-cross-android ./configure 
should do the trick now for you. that was really a bad thing that I had not implemented that.

fornwall:
> I don't really see the necessity to remove rindex, as I highly doubt it will be removed in the next 10 years, deprecated or not.

Well, it's already unavailable on the Android libc and causes a small complexity here with the configure and ifdef checks. I guess the question is what is gained by using rindex() - is there any advantage at all (any platform supporting that but not strchr()) in keeping it? Anyway, thanks for the quick fix!

FYI, I just packaged Hydra v8.4 for [Termux](https://termux.com/), so users there can install hydra with a simple `apt update && apt upgrade`.

vanhauser-thc:
the thing is - I want hydra to be able to compile on all circumstances. from the old old times of hydra it could even compile on ultrix, informix and other uncommen platforms.
now with so many changes in the code that it might not compile on old machine anymore and that old unix systems are absolutely rare that primary requirement is not very helpful anymore.
still I try to have backward compability where possible.

PR #196, opened 2017-03-01
New feature: output results in JSON format.

veggiespam:
# New Feature: export results as JSON

We have a need to parse the output of Hydra to flag issues for clients.  I have a script that does various cut/sed/perl to extract each bad user account, but a system that allows user IDs with spaces in names just made this more difficult.  So, I found it easier to modify Hydra to simply output this information in JSON.  This also simplifies our downstream processing of the data.  So, I am contributing this back to the community.

I added a new `-b` command line flag which requires a single parameter of `text`, `json`, or `jsonv1` to determine the format of results file.  Choosing `text`, the default, will output the results file exactly as it is now. However, `json` or `jsonv1`, will output the same information but in JSON format.  The options:

* `text` - output as it is now
* `json` - the latest version of the JSON output
* `jsonv1` - version 1.0, 1.1, 1.x, of the JSON output

Future options might include:

* `jsonv2` - version 2.0 of the output with some compatibility-breaking formatting changes that users have requested
* `xmlv1` - XML version 1.0
* `zap` - something which might format data to be fed into ZAP Proxy
* etc... 

In order to replicate the current data output to the results file, this is the JSON format I created (with pretty-print applied):

```
{
    "errormessages": [],
    "generator": {
        "built": "2017-03-01 14:44:22",
        "commandline": "hydra -b json -o results.log ... ...",
        "jsonoutputversion": 1.0,
        "server": "127.0.0.1",
        "service": "http-post-form",
        "software": "Hydra",
        "version": "v8.5-dev"
    },
    "quantityfound": 2,
    "results": [
        {
            "host": "127.0.0.1",
            "login": "bill@example.com",
            "password": "bill",
            "port": 9999,
            "service": "http-post-form"
        },
        {
            "host": "127.0.0.1",
            "login": "joe@example.com",
            "password": "joe",
            "port": 9999,
            "service": "http-post-form"
        }
    ],
    "status": "success"
}
```

`errormessages` is an array of strings would be the various items printed to stderr at the end of the Hydra's run.  `status` is either the text `success` or `errors` depending on completion.  

There probably is more needed for a more thorough export of the results, but I believe this is a good first start.  The code might need some munging to get it into your style. Nevertheless, I'd like to see this added to Hydra v8.5-final as a "beta" feature.

Thanks.

### DIFF
--- a/hydra.c
+++ b/hydra.c
@@ -255,6 +255,7 @@ typedef struct {
   int exit_found;
   int max_use;
   int cidr;
+  int outfile_format;            // 0 = plain text, 1 = JSONv1, [future --> ]  2 = JSONv2, 3=XMLv1, 4=...
   char *login;
   char *loginfile;
   char *pass;
@@ -357,6 +358,8 @@ void help(int ext) {
   printf("  -M FILE   list of servers to attack, one entry per line, ':' to specify port\n");
   if (ext)
     printf("  -o FILE   write found login/password pairs to FILE instead of stdout\n");
+  if (ext)
+    printf("  -b FORMAT broker -o FILEs in (text[default], json, jsonv1) format\n");
   if (ext)
     printf("  -f / -F   exit when a login/pass pair is found (-M: -f per host, -F global)\n");
   printf("  -t TASKS  run TASKS number of connects in parallel per target (default: %d)\n", TASKS);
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
thanks for the patch!

Issue #27, opened 2014-12-04
hydra -R issue


cporichie:
Hi,

Using the v8.1-dev most recent pull, when trying to run the restore file hydra -R, I am getting the following error.

**\* Error in `hydra': double free or corruption (out): 0x00007fb491544260 ***

Here is the original command with a few things obviscated. I tried various command adjustments with no luck.

hydra -v -l user -x '8:8:aA1' -f -F -t 1 -m '/page.php:username=^USER^&password=^PASS^:Incorrect' 10.0.0.15 http-post-form -o /home/user/Desktop/mso.txt -w 10

Any suggestions or do you need anything else from me?

vanhauser-thc:
I can not reproduce the bug and cant see where it might come from.
can you check if the same bug is present with ftp? you can use as target the following:
  host: www.mh-sec.de    user: foo

ihatetoregister:
I'm not sure this is related but I get tons of the following error message when trying to restore a ssh session:

    hydra(86236,0x7fff78000300) malloc: *** error for object 0xd00007ff7d0c06b2: pointer being freed was not allocated
    *** set a breakpoint in malloc_error_break to debug

The session was started with command:

    hydra -t 4 -l root -P passwords.txt ssh://10.0.1.3

I installed the lastest (I think) hydra using homebrew:

    brew install --with-libssh --HEAD

vanhauser-thc:
what is the hydra version in brew? if it is not 8.2-dev it is not helping :)

ihatetoregister:
Homebrew seem to clone the github repo and just typing the command hydra shows "v8.2-dev" so it seem to be that one yes. However the resume behavior was the same with v8.1.

dkorzhevin:
Hi!

I have same crash with hydra 8.1 in latest release of Kali Linux. It is 100% reproducible.

I even record short video about crash:

https://www.youtube.com/watch?v=ybAejmCQYzw

I will be grad to follow any instructions to debug and help fix this bug.

vanhauser-thc:
@dkorzhevin - a youtube video does not help anything, sorry.
is your crash on 32bit or 64bit Linux?
If it is 32bit, please upload the hydra binary and a core file ("ulimit -c 20000000" before running hydra)

dkorzhevin:
This is:
# cat /etc/issue
Kali GNU/Linux 1.1.0 \n \l

# uname -a
[...]

PR #209, opened 2017-04-24
do not assume sizeof(int) is the pointer size

lukas227:
sizeof(int) is not necessarily the same as the pointer size. On my amd64
system sizeof(int) is 4 bytes while the pointer size is 8 bytes. The
change solves https://github.com/vanhauser-thc/thc-hydra/issues/27 for
me.


vanhauser-thc:
thanks!

PR #217, opened 2017-05-20
Bugfix for issue 192 -- increased URL/POST/cookie data size to 6096 b…

petrock6:
…ytes from 1000 bytes.


vanhauser-thc:
there was an off-by-one overflow which I fixed. I also put down the string size to 6000 bytes just to be sure that we have not an overflow anywhere. I hope I did not overlook something.

thanks for the patch!

PR #220, opened 2017-06-01
Using var for target instead of indexing

Diadlo:


vanhauser-thc:
Hi,

this change makes the code not easier readable nor reduces it the lines of code.
instead it is a big change which can have issues if a single line is wrong.
so this one I rather dont want to merge, sorry.

Diadlo:
It's make code easier to read a lot. Just believe to man, who try to read this :)
Would it be suited to you, if I will split this PR on smaller ones with the same logic of changes?

PR #228, opened 2017-06-10
Useless check

Diadlo:


vanhauser-thc:
maybe I havent looked at my code for a very long time - why do you thing the if (sock >= 0) close(sock); is useless?

Diadlo:
@vanhauser-thc Because of [this](https://github.com/vanhauser-thc/thc-hydra/blob/master/hydra-pop3.c#L519)

vanhauser-thc:
this is a good example how a upstream patch should not be :)
it changes not one but several things, and the explanation is not helpful.

so I have to lookup every change in the file by hand to assess if it is OK or not.

please for future patches: they should fix one type of thing only. the description should make clear what it does (unless it is abolutely obvious)

here I don't understand your chance to sasl.c:
-  } while ((pbuffer[0] != '\0') && (pbuffer[0] > 31) && (ind < array_size));
+  } while ((pbuffer[0] > 31) && (ind < array_size));

why did you remove the pbuffer[0] null check?

Diadlo:
I edited the commit messages

vanhauser-thc:
thanks!

PR #230, opened 2017-06-11
Maxfail

Diadlo:


vanhauser-thc:
This can't work.
you define int s = success < 5 ? 6 - success : 1; but are not using it, instead you use b in the next line which you did not set.
and it looks like you are increasing the overall maxfail number by this. this is actually a bad idea as it would result in more connect attempts per fork and more load on the server.
the idea behind this is to automatically reduce the number of parallel connections until most of them get through without problems.

Diadlo:
@vanhauser-thc Sorry, I've renamed var after last compile test. Fixed
I don't change the logic at all. I just extract huge calculation in one variable

Diadlo:
I've split commits on smaller to make changes easier to understand

vanhauser-thc:
thanks

PR #233, opened 2017-06-22
Service include

Diadlo:


vanhauser-thc:
What I like about the patch:
moving the help for a module to the module is great and makes hydra.c smaller

What could be better:
every service should have a help stub, even if it is empty.

What I don't like:
the services/ subdirectory
the service.h files

I understand why you did that, but I do not like it that way.

In the future it would be beneficial if you write to me first for changes that you would like to do, otherwise you might waste your time.

And finally: makeing hydra better (like fixing vnc, smbv2+v3, rdp fixing etc.) is more important than making the code easier to read.

Diadlo:
@vanhauser-thc I thought about help stub, I can do it later
Why you don't like `services/` and `service.h`?
How can I contact with you?

I start hydra contributing because I wanted to fix #27, but I just couldn't read this code. Maybe it's my bad. But if someone who want to contribute can't understand the code, he can't contribute. If you still think, that making the code easier to read is not so important I will stop it.

Diadlo:
Also, speed of feature development and bug fixes depends of code quality

vanhauser-thc:
as I said there are other areas where help/time is more important and gave examples.
my email is everywhere, e.g. in the README: vh(at)thc.org

vanhauser-thc:
I appreciate you investing your time.
but changing the layout of a lot of stuff I am used to work with (as bad as it might be) - I am the one who has to work with it most of the time - is not something that I want for the code. e.g. the services/ directory, the services.h or the other patch you did with the (target*) target = targets one
rather help smb to get v2 and v3 support with libsmb, check what needs to be fixed so the rdp module works again, and it seems that the mysql is not always working when it should (client library problem? mysql server too new?)

Diadlo:
Did you read my comment? It's very hard to contribute if you can't understand the logic of code. I understand, that you can easily read the code you wrote, but it may be not so obvious for another people.

Diadlo:
Ok. I will stop

PR #235, opened 2017-06-24
Reset redirect flag and fix redirect port error issue

mindon:
fix following 2 issues:

1. when fail is a redirect url, matched is not - but the redirect flag is still 1 - causing match missed

2. when redirect to a relative url on a port not 80, it fails to attach the port to the redirect url


vanhauser-thc:
thanks!

PR #244, opened 2017-08-01
add radmin2 to xhydra selection and man page

lukas227:
The two commits add radmin2 to both the xhydra selection and the hydra man page.  I've also dropped the listing of modules from the xhydra man page to reduce the duplication and made a few minor markup changes.

Thanks for always being open to minor changes (and improving the spelling even more after the last PR)!


vanhauser-thc:
thanks!

PR #250, opened 2017-08-17
Clang fix

crondaemon:
All the commits come from my attempt to compile hydra with clang. It works, but many annoying warnings come along.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-CC?=gcc
+CC=clang
 STRIP=strip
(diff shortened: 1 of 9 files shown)


vanhauser-thc:
Hi,

thanks for the patch.

I used some of the changes and applied them by hand.
the majority however does not fit: I prefer having functions use unsigned char* instead of signed char*. too often had I had a bug because of that.

So if you want to get rid of warnings, thats fine with me, as long as you do not reassign functions or variables to signed char.
rather type cast a char to unsigned char or redefine a variable to unsigned char and char to char* where this is expected for c functions (like send, recv, etc.).

thanks!

PR #268, opened 2017-10-05
fix build with musl libc

ebfe:
```
cc -I. -O3     -o pw-inspector  pw-inspector.c
pw-inspector.c:39:1: error: unknown type name 'int32_t'
 int32_t main(int32_t argc, char *argv[]) {
 ^~~~~~~
pw-inspector.c:39:14: error: unknown type name 'int32_t'
 int32_t main(int32_t argc, char *argv[]) {
              ^~~~~~~
make: [Makefile:70: pw-inspector] Error 1 (ignored)

Now type make install

```

### DIFF
--- a/pw-inspector.c
+++ b/pw-inspector.c
@@ -3,6 +3,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <ctype.h>
+#include <stdint.h>
 
 #define PROGRAM		"PW-Inspector"
 #define VERSION		"v0.2"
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
thanks but the correct fix is actually to have int main(int argc, not int32_t.
that is what you get when you do global replacements with sed ;)
I will fix it now, had not noticed it

ebfe:
> thanks but the correct fix is actually to have int main(int argc, not int32_t.

That was part of the patch.  And stdint.h is needed for the later usages of int32_t.

PR #269, opened 2017-10-05
fix build with musl libc

ebfe:
0a50b41be99f1b2fe2b1c2235a3a0a18c2d92f36 only solves half of the problem.

```
cc -I. -O3 -pie -fPIE -fstack-protector-all --param ssp-buffer-size=4 -D_FORTIFY_SOURCE=2 -Wl,-z,now -Wl,-z,relro    -o pw-inspector  pw-inspector.c
pw-inspector.c: In function 'main':
pw-inspector.c:40:3: error: unknown type name 'int32_t'
   int32_t i, j, k;
   ^~~~~~~
[...]
```

### DIFF
--- a/pw-inspector.c
+++ b/pw-inspector.c
@@ -3,6 +3,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <ctype.h>
+#include <stdint.h>
 
 #define PROGRAM		"PW-Inspector"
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
you are right, I rushed it as I have very little time. which cost more time in the end ;)
thanks!

Issue #267, opened 2017-10-02
http-get-form does not send absolute URL with an http proxy


aayla-secura:
[HTTP RFC](https://tools.ietf.org/html/rfc2068) says absolute URLS must be used with proxies. Hydra correctly implements this with http-post-form, but not with http-get-form, resulting in 400 Bad request from the proxy. Tested with v8.6. Example with dvwa:

```
HYDRA_PROXY_HTTP='http://127.0.0.1:8080' hydra -d -l admin -p password 192.168.56.102 http-get-form "/dvwa/vulnerabilities/brute/:username=^USER^&password=^PASS^&Login=Login:S=Welcome"
```

shows the first packet not including any query parameters and correctly using an absolute URL, gets a 200 OK with the login page. Next packet sends the query parameters but issues a GET /dvwa/... and receives a 400 Bad request from the proxy (mitmproxy).

vanhauser-thc:
yeah that looks wrong.
its not my code and I am swamped with work, but I will try to fix it on the weekend or next week.

or you can try for a patch if you want.

PR #270, opened 2017-10-05
Fixes issue #267

aayla-secura:
Fixes issue [#267](https://github.com/vanhauser-thc/thc-hydra/issues/267): http-get-form does not send absolute URL with an http proxy


vanhauser-thc:
good job, thank you!

PR #294, opened 2017-12-21
Don't add extra slash to url if exist

mostafahussein:
This PR removes an extra `/` which appears in the current hydra's output like this:
```
[DATA] attacking http-get://foo.com:80//bar/
```
And change it to be like this:
```
[DATA] attacking http-get://foo.com:80/bar/
```

The `hydra_options.miscptr` parameter will be checked if it starts with a slash then no need to add another one. If you would like to fix the issue in another way please let me know so I can fix it in an appropriate way.

The command used during testing:
```
hydra -l user -P mylist http-get://foo.com/bar/
```


vanhauser-thc:
thanks!

Issue #100, opened 2016-03-12
Problem with http-get


xPapla:
Hello,
I have a problem whenever I try to use http-get testing my current router.
I tried:

``` bash
hydra -l user -p user 192.168.55.1 -m "/" http-get -d > hydra_logs.txt 2>&1
```

 but it finished with 

> 1 of 1 target completed, 0 valid passwords found

To be sure that everything is okay with login and password I also tried:

``` bash
wget --user=user --password=user http://192.168.55.1/ -d -o wget_logs.txt
```

and it has succeeded.

Full logs:
- [wget_logs.txt](https://github.com/vanhauser-thc/thc-hydra/files/170459/wget_logs.txt)
- [hydra_logs.txt](https://github.com/vanhauser-thc/thc-hydra/files/170460/hydra_logs.txt)

also a result of successful wget:
- [index.html](https://github.com/vanhauser-thc/thc-hydra/files/170462/index.html.txt)

vanhauser-thc:
can you please check out the current repository state and see if it works?

if it does not, it is because your router is behaving non-standard, it seems to require a cookie: present.

jamieOH:
I am also having issues and think it may be due to a cookie - but I haven't found any solid information of the syntax of how to add a cookie. Can you maybe link me a resource or tell me the basics?

CookieMichal:
So, I cloned eb3628d and tried again, but still hydra fails - [full logs](http://termbin.com/z1j8)

vanhauser-thc:
hi guys,

sorry, http-get does not support cookies at the moments - simply because they are not necessary.
it is the router that is non-standard.
the cookie functionality is present in the http-form module, so I (or somebody else) might port it at some point to support also this non-standard behaviour.
I currently do not have the time for it, I will put it on the TODO list though.
If one of you guy has the time to do it, that would be great.

PR #342, opened 2018-07-16
Optionally add request headers to http-get

### LINKED ISSUE #321: Cookies/request headers with http-get
I'm trying to use Hydra to test HTTP basic auth credentials. The system in question will only handle this correctly if a fixed cookie is included in the request along with the HTTP basic auth credentials.

I don't see a way to add cookies or custom request headers when using the http-get module, only the form and POST-related modules. Am I misreading the help? I don't want to send the username and password as URL or body parameters, which is why I'm using http-get.

Additionally, the behaviour of the target system is that when the cookie is *not* present, it sends an HTTP 403 response. Hydra is interpreting this as if the password is valid, because usually a 403 will be returned if authentication is successful, but authorization is not. Is there any chance of making it a command-line option to treat 403, 404, etc. as failed auth? Right now Hydra is telling me that all of the passwords in my list are valid, which is not the case.

mathewmarcus:
This adds support for optionally adding headers to `http-get` and `http-post` via `(h|H)=My-Hdr\: foo`. I believe this should close #321 and #100. Please let me know if I've made any mistakes or if there's anything you'd like to see changed. Thanks!

### DIFF
--- a/hydra-http-form.c
+++ b/hydra-http-form.c
@@ -50,6 +50,7 @@ Added fail or success condition, getting cookies, and allow 5 redirections by da
 */
 
 #include "hydra-mod.h"
+#include "hydra-http.h"
 
 /*	HTTP Header Types	*/
 #define HEADER_TYPE_USERHEADER                  'h'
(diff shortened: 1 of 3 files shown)


vanhauser-thc:
Hi,

thanks for the patch!
did you check that http-form is still working? and http-get too if the cookies/header etc. are not used?

mathewmarcus:
Yes I tested several scenarios using both http-form-post and http-get with and without headers. That said I would welcome/appreciate any additional testing you or anyone else would like to do.

vanhauser-thc:
If you find the time to add the C= option as well this would be great as several wifi routers use basic authentication with set-cookie when submitting the credentials. Also learning the cookies (code is also in http-form) would be beneficial

Issue #347, opened 2018-07-27
http-form-post cookie not set from POST response redirect


mathewmarcus:
From what I can see, hydra is currently not setting cookie(s) returned from an http-post-form request in the `Set-Cookie` header. Specifically, if the POST response is a 300-level redirect, the cookies are not passed along in the subsequent GET request. Here is an example command:

```bash
hydra -l foobar -p barfoo 192.168.1.7 http-post-form '/cgi-bin/courierwebadmin.cgi:loginpass=^PASS^&do.login=1&login=Login:ERROR'
```

This is the POST response

```
HTTP/1.1 302 Found
Date: Fri, 27 Jul 2018 14:03:59 GMT
Server: Apache/2.2.14 (Ubuntu) mod_mono/2.4.3 PHP/5.3.2-1ubuntu4.30 with Suhosin-Patch proxy_html/3.0.1 mod_python/3.3.1 Python/2.6.5 mod_ssl/2.2.14 OpenSSL/0.9.8k Phusion_Passenger/4.0.38 mod_perl/2.0.4 Perl/v5.10.1
Cache-Control: no-store
Pragma: no-cache
Set-Cookie: courierwebadmin=barfoo; path=/
Location: http://192.168.1.7/cgi-bin/courierwebadmin.cgi
Content-Length: 0
Connection: close
Content-Type: text/plain
```

And here is the subsequent GET request

```
GET http://192.168.1.7:80/cgi-bin/courierwebadmin.cgi HTTP/1.0
User-Agent: Mozilla/5.0 (Hydra Proxy)
Content-Length: 0
Content-Type: application/x-www-form-urlencoded
Cookie: 
Host: 192.168.1.7
```

This is an issue because the error string - in this case `ERROR` - is only displayed if the correct cookie is present.

PR #348, opened 2018-07-27
set cookie for http-form-post redirects

mathewmarcus:
From the tests I've conducted, this seems to fix issue #347.  Please let me know if I've made any mistakes or if there's anything you'd like to see changed. Thanks!


mathewmarcus:
I see it as useful to other people. Moreover I think it is the expected behavior, i.e. sending cookies along with additional requests made to the same domain. For the case I described in issue #347, omitting this cookie would result in a false positive.

vanhauser-thc:
thanks!

mathewmarcus:
Thank you!

Issue #357, opened 2018-09-29
Each printf digit expands to take two places


deargle:
```
uname -a
Linux kali 4.18.0-kali1-686-pae #1 SMP Debian 4.18.6-1kali1 (2018-09-10) i686 GNU/Linux
```
`Hydra v8.7-dev (c) 2018` (I'm on master branch, pulled today)

Hydra verbosity logs are borked for a head-scratching reason. Each `printf` digit expansions from the hydra functions is taking _two_ positions. 

An odd example: I added this in `hydra.c:3460`, where `math2=896525`

```
printf("foo %lu bar %lu foobar %lu %lu, math2, math2)
```
Output:

```
foo 896525 bar 0 foobar 896525 0
```

This is a problem because the verbose logs end up looking like this:

```
[ATTEMPT] target <target> - login "<login>" - pass "123456" - 1 of 0 [child 14344399] (0/0)
[ATTEMPT] target <target> - login "<login>" - pass "12345" - 2 of 0 [child 14344399] (0/1)
```

...when they should look like:

```
[ATTEMPT] target <target> - login "<login>" - pass "123456" - 1 of 14344399 [child 0] (0/0)
[ATTEMPT] target <target> - login "<login>" - pass "12345" - 2 of 14344399 [child 1] (0/0)
```

I'm novice in c, so I'm banging my head. Perhaps something to do with output buffering that hydra is doing? Scratching my head.

deargle:
Seems to fix it if all `%lu`'s are changed to `%llu`.

dmaciejak:
Should be fixed now, please test.

PR #358, opened 2018-09-29
change long unsigned to long long unsigned

deargle:
Fixes for i686. I admittedly have no idea what the impact of this change would be on Kali x64 systems, but PRing so that you can look if you want to...

See #357

### DIFF
--- a/hydra-ssh.c
+++ b/hydra-ssh.c
@@ -33,6 +33,7 @@ int32_t start_ssh(int32_t s, char *ip, int32_t port, unsigned char options, char
   if (strlen(pass = hydra_get_next_password()) == 0)
     pass = empty;
 
+
   if (new_session) {
     if (session) {
       ssh_disconnect(session);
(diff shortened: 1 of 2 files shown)


deargle:
I tested on x64, works fine from what I can tell.

`Linux kali 4.17.0-kali1-amd64 #1 SMP Debian 4.17.8-1kali1 (2018-07-24) x86_64 GNU/Linux`

vanhauser-thc:
Hi, 

the issue with the long long int patch is that it is architecture dependant if its "long long int" or "int" - 64 or 32 bit.
so this patch would just switch the warnings to the other platform.
I think I will switch to #define as this is the only way I see this can work.

deargle:
Sounds good. I am an ultra novice in C, I defer to you. Although I would
note that I never got a warning. Regardless, I'm happy to test whatever you
want to do if you'd like that.

dmaciejak:
Please check the last commit, it's fixing some compilation warning for long unsigned printing that depend on the artchitecture https://github.com/vanhauser-thc/thc-hydra/commit/e1e708d1a1b1b96849e21fb4e9a45fef0e85dfab

PR #394, opened 2019-03-22
Fix compilation error when Mongo and Memcache libraries are absent

mlsendian:
Tiny change, there's a duplicated function name: dummy_mcached.


vanhauser-thc:
thanks for spotting and fixing!

@dmaciejak - seems like you forgot something ;)

dmaciejak:
oops thanks !

Issue #390, opened 2019-03-03
rdp module does not support the current protocol


sirty69:
[ERROR] the rdp module does not support the current protocol, hence it is disabled. If you want to add it, please contact vh@thc.org

What's the problem ?

vanhauser-thc:
that someone has to write the module to support the current protocol.
I dont have the time for that.

baotang2118:
> [ERROR] the rdp module does not support the current protocol, hence it is disabled. If you want to add it, please contact [vh@thc.org](mailto:vh@thc.org)
> 
> What's the problem ?
@sirty69 @vanhauser-thc 
I found the answer for this topic even though it's not the official one. When Hydra hit the right user + password, it sit there for a while then quit program, it doesn't continue bruteforcing. This problem happens seem the server suddenly shutdown the connection and Hydra has to wait for connection time out. Although the output is:
`[STATUS] 1.00 tries/min, 5 tries in 00:00h, 0 to do in 03:00h, 1 active`
`[STATUS] 0.83 tries/min, 5 tries in 00:00h, 0 to do in 03:00h, 1 active`
`[STATUS] 0.56 tries/min, 5 tries in 00:00h, 0 to do in 03:00h, 1 active`
`[STATUS] 0.36 tries/min, 5 tries in 00:00h, 0 to do in 03:00h, 1 active`
`[STATUS] 0.33 tries/min, 5 tries in 00:00h, 0 to do in 03:00h, 1 active`
Press Ctrl+C. The last password that Hydra tried may be the correct password. I tried, it work for me. It only works for 1 username. I tried more than 1 username, then the progress stops because of so many attempt. Here is my script:
`hydra -t 1 -V -f -l administrator -P wordlist.txt rdp://my_server_ip`
Sometime you will see `1 of 1 target successfully completed, 0 valid password found`

petercunha:
Bump. An RDP module would be very nice.

petercunha:
thc is busy I suppose. I will do my best to code this and pull request.

God bless THC.

vanhauser-thc:
I patch was submitted which uses libfreerdp.
Everyone please sync your github thc-hydra, configure, make, make install, test it and report!

vanhauser-thc:
thanks a lot to @loianhtuan !

zordis:
[ERROR] Compiled without FREERDP2 support, module not available!

How compiling with FREERDP2 support?

loianhtuan:
[...]

PR #405, opened 2019-04-23
replace rdesktop with freerdp for rdp module

loianhtuan:
to build we need FreeRDP 2.0 from https://github.com/FreeRDP/FreeRDP
It should solve issue #390


vanhauser-thc:
thanks a lot!
I can not test it, but will try to crowdsource it :)

PR #408, opened 2019-04-25
rebuild JSON output file in creating output file

Galaxy-cst:
The a+ mode will cause JSON format error. It is a better choice JSON format file does not support appending.


vanhauser-thc:
Makes sense, however this will result in unwanted behaviour if session restore -R is being used. I will accept the pull request though.
Thanks!

PR #409, opened 2019-04-27
Rebuild JSON report file in first time and append while using -R

Galaxy-cst:
Hi,
My last PR did ignore the -R situation. This time I fixed this problem.
Now the JSON output file will be rebuilt when the session is first started, and in the -R it append.
I have tested that the JSON output is correct whether it's new or restored.

PS:
I try to make this clean, so I add a new "outfile_open_type" to avoid repeated "fopen".

Thanks & Regards,
Galaxy


vanhauser-thc:
perfect, thanks!

PR #414, opened 2019-05-12
fix NTLM authentication

rmaksimov:
There is an issue with NTLM authentication.
First of all, while using an HTTPS module, Hydra sometimes results in false negative or freezes when receiving data from a socket. So, an additional check is necessary in a while loop (hydra_receive_line function):
```
#ifdef LIBOPENSSL
       || use_ssl && SSL_pending(ssl)
#endif
```
Second, there should be keep-alive value of Connection HTTP header (or it may be omitted since this is default behavior in HTTP/1.1) when NTLM authentication is used.

In addition, hydra_receive_line function has been rewritten, so it doesn't have a duplicate code now. I have tested it only for HTTP/HTTPS Basic and NTLM authentication types, but not for the other modules.

### DIFF
--- a/hydra-http.c
+++ b/hydra-http.c
@@ -70,8 +70,8 @@ int32_t start_http(int32_t s, char *ip, int32_t port, unsigned char options, cha
       fooptr = buffer2;
       sasl_digest_md5(fooptr, login, pass, buffer, miscptr, type, webtarget, webport, header);
       if (fooptr == NULL) {
-	free(buffer);
-	free(header);
+        free(buffer);
+        free(header);
         return 3;
       }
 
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
thank you for the big fix.

question though - did you check that normal http auth is still working?

Also I see this nonsense that was in before:
"Connection: close\r\n...\r\nConnection: keep-alive"
in your change neither is present - wouldnt one of these be needed there?

thanks!

rmaksimov:
> question though - did you check that normal http auth is still working?

I have tested Basic auth and it works fine. One thing I didn't test is authentication via proxy.

> Also I see this nonsense that was in before:
"Connection: close\r\n...\r\nConnection: keep-alive"
in your change neither is present - wouldnt one of these be needed there?

NTLM authentication requires keep-alive value, but it may be omitted since this is default behavior in HTTP/1.1

vanhauser-thc:
ok thanks!

PR #415, opened 2019-05-12
add authentication type switching option

rmaksimov:
If a server responds with multiple authentication type headers
```
WWW-Authenticate: NTLM ...
WWW-Authenticate: Negotiate
WWW-Authenticate: Basic realm="..."
```
then there is a need for an option for switching authentication type to test credentials using the certain type instead of Basic

### DIFF
--- a/hydra-http-form.c
+++ b/hydra-http-form.c
@@ -50,6 +50,7 @@ Added fail or success condition, getting cookies, and allow 5 redirections by da
 */
 
 #include "hydra-http.h"
+#include "sasl.h"
 
 extern char *HYDRA_EXIT;
 char *buf;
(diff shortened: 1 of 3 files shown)


vanhauser-thc:
I am really sorry for this - this is the wrong way.
You are adding a command line option to hydra which is just for a specific module. This should never be the case. The "right" way is to use -m MODULE_OPTION_STUFF values.
so for example the miscptr (thats how it is called in hydra-http.c) could check for a ':' char, and after this one check for MD5, NTLM etc. (and of course put a null byte where the ':' was to the URL parameter is stil lintact).
This is how it is done in e.g. hydra-pop3, hydra-imap etc.

rmaksimov:
Thank you for feedback.
I haven't used advanced options of modules before, so the previous fix was really wrong.
I have made the necessary changes and everything is ok now.

vanhauser-thc:
you put it into the wrong module  :)
you put it into hydra-http-form instead of hydra-http where you want/need it.

for hydra-http the miscptr will contain the uri that is tested (e.g. /secret/) so that it contains an extra parameter you need a seperator, for which we use the ':' character.
and the selector of the authentication type should then be a string, e.g. "BASIC", "NTLM", etc.
and after your parsing you must truncate the string at the seperator so that the uri can be used.

so: *miscptr would contain "/secret/:NTLM"
after your parsing it may only contain "/secret/" anymore.

rmaksimov:
I put it there because there is parse_options function call from hydra-http.c that is described in hydra-http-form.c. Moreover, there is nothing that parse 'h' and 'H' option in hydra-http.c, it also parses using function from hydra-http-form.c, isn't it? So I thought parse_options function is like a helper function used to parse options for these modules.

>  and the selector of the authentication type should then be a string, e.g. "BASIC", "NTLM", etc.
and after your parsing you must truncate the string at the seperator so that the uri can be used.
>
>so: *miscptr would contain "/secret/:NTLM"
after your parsing it may only contain "/secret/" anymore.

Yeah, I saw it in the modules you mentioned in the first reply, but the strange behavior described above misled me.

vanhauser-thc:
oh man I am so sorry.

I thought you put the parse_options into http-form. but you didnt you just used what was already there.
no clue who started it but its not how it should be ... but well ...

because I thought I did it I thought I just quickly implement it myself which is easier than explaining .... started that, and while doing that step by step I understood that you did exactly what I just didl.

so sorry man ... my bad. that what happens when I accept lots of patches from lots of people - I sometimes forget or dont see the weirdness that creeps into the code.

PR #417, opened 2019-05-15
Header value parsing fix and some misc changes

rmaksimov:
There was a mistake in pointer offset value in advanced options of HTTP/HTTPS modules.
Also, the description of the option and the format of the output information were fixed.


vanhauser-thc:
thanks!

PR #418, opened 2019-05-16
fix typo and the offset value of miscptr

rmaksimov:
There should be an offset in the pointer value for everything to work properly


vanhauser-thc:
thanks for fixing my fuckup

vanhauser-thc:
can you please test that the http module works with the A=BASIC etc. option? So its doing the right thing?

rmaksimov:
Have tested Basic and NTLM, both work fine

vanhauser-thc:
thank you!

PR #419, opened 2019-05-18
Worked around APR_PATH_MAX errors on some Linuxes when SVN is enabled.

hlein:
On Gentoo Linux (and possibly others?) with Subversion 1.12.0,
compilation of hydra-svn.c fails with:

In file included from /usr/include/subversion-1/svn_client.h:34,
                 from hydra-svn.c:9:
/usr/include/apr-1/apr.h:632:2: error: #error no decision has been made on APR_PATH_MAX for your platform
 #error no decision has been made on APR_PATH_MAX for your platform
  ^~~~~

This happens when PATH_MAX is not defined.

PATH_MAX is defined by /usr/include/linux/limits.h, but rather than
include'ing that directly and possibly breaking other platforms,
include sys/param.h (which will include linux/limits.h indirectly)
iff PATH_MAX is not defined and sys/param.h exists.

I based the approach on how math.h is handled.

### DIFF
--- a/configure
+++ b/configure
@@ -557,6 +557,10 @@ for i in $INCDIRS ; do
         fi
     fi  
 done
+SYS_PARAM=""
+if [ -f "$SDK_PATH/usr/include/sys/param.h" ]; then
+  SYS_PARAM=-DHAVE_SYS_PARAM_H
+fi
 if [ "X" != "X$DEBUG" ]; then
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
thanks!

hlein:
That was quick!  Thank you!

PR #432, opened 2019-07-24
HTTP-GET add end condition. Simulary like in http-form

poucz:


vanhauser-thc:
hi,

thanks for the pull request. several issues though.

1.                  hydra_report(stderr, "End condition is %s, mod is %d\n",end_condition,end_condition_type);
this will be printed by every forked head. not good. put an  if (debug)   

2. did you test the code? from looking at it I am not sure ... ...
```
-    hydra_report_found_host(port, ip, "www", fp);
-    hydra_completed_pair_found();
+    if(end_condition_type>=0 && hydra_string_match(http_buf,end_condition)!=end_condition_type){
+		if (debug)
 +                       hydra_report(stderr, "End condition not match continue.\n");
+		hydra_completed_pair();
+	}else{
 +           hydra_report(stderr, "END condition %s match.\n",end_condition);
+	    hydra_report_found_host(port, ip, "www", fp);
+	    hydra_completed_pair_found();
+	}
```

this might work if F= and S= is used - but if end_condition_type == -1 then the module will not report finding a valid password ... 

please. submit. proper. patches.

poucz:
Hi,

1) Ok I will apply it.
2)I don't think so.I think, if "end_condition_type == -1" then program behavior is same like without my patch.....
Please look to attachment,  It's diff code, and I mark yellow color if "end_condition_type == -1".
screen



Maybe I'm wrong, please let me know.

vanhauser-thc:
you are right, I was kinda unable to read the code properly ... by bad ... I blame the heat :)

vanhauser-thc:
I will accept the pull request and then squash typos you put in ;)

vanhauser-thc:
also if you want you can also port the C= option to get cookies (needed for some wifi routers)

Issue #316, opened 2018-04-10
The program is pause when the folder is read


src-kun:
hydra -C /tmp ftp://192.168.0.1

The program is pause when the folder is read.

vanhauser-thc:
yeah thats dumb, but so non-critical I am not sure I will spend the little time I have on this.
you can submit a patch though.

src-kun:
kk

PR #443, opened 2019-08-30
Fix dump with folder

loonydev:
Hi, I know that it's looks like a crunch, but when user send `-C /tmp` hydra start endless loop in hydra.c:L1054-L1090. Add flag which trigered if nothing readed.

Screen Shot 2019-08-30 at 3 00 33 PM

And good case 

Screen Shot 2019-08-30 at 3 00 53 PM

Close https://github.com/vanhauser-thc/thc-hydra/issues/316


vanhauser-thc:
thank you!

PR #467, opened 2019-10-21
http-post 

e2002e:
Hi / Guten abend.
As you may have noticed the http-post/get-form is broken.
I noticed it can be fixed by simply rolling back, but the comments seem to specify that it has to be written using strtok(), so that's what i've done.
Right now I can certify it's working BUT I have not tested nor made much investigation about the escapes inside the miscptr ( -m option), it's managed already for escaping columns ":"  and as I type I don't see which other characters would need to be escaped. Hoping you pull this. Aufvieder sehen. :P

### DIFF
--- a/bfg.c
+++ b/bfg.c
@@ -192,12 +192,12 @@ uint64_t bf_get_pcount() {
 
 int accu(int value)
 {
-	int sum = 0;
-	for(int i=1; i<=value; ++i)
-	{
-		sum+=i;
-	}
-	return sum;
+  int i = 0;
+  for(int a=1; a<=value; ++a)
+  {
+  	i+=a;
+  }
+  return i;
 }
 
 char *bf_next(_Bool rainy) {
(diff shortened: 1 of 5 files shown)


e2002e:
i had opened this a few weeks ago, it's the fix for the http-post-form

e2002e:
hey @vanhauser-thc, regarding the issue  #467, Here there are no changes compared to v8.8 in the way post parameters are parsed, this is just a roll-back, with the checks for escapes before colons, and the use of hydra_strrep in case we have these. All of this is in the function ptr_header_node initialize at line 1213 in the file hydra-http-form.c.
This works and there should not be issues. Cheers

vanhauser-thc:
I understand what you were trying to do, but besides fixing a new bug this change introduced, it reintroduces a bug that this fixed :)
I tried something different, its checked in, please test

Issue #197, opened 2017-03-02
SMBv2/3 support


hatlord:
Hi,

Is there any possibility of SMBv2/3 support within Hydra? We recently have seen SMBv1 being disabled after advise given by CERT following on from the Shadow Brokers announcement of a possible SMB exploit: 
https://www.us-cert.gov/ncas/current-activity/2017/01/16/SMB-Security-Best-Practices

Not sure how commonplace this is going to be yet, but I figured it was worth asking!

Thanks,

vanhauser-thc:
that would be a good thing I agree.

smbv1 has orginally been implemented by previous Nessus developer, not me. and it was done with just calling crypto libraries.

SMBv2+v3 are more complex and would require using an external smb library instead.
I do not have time for that at the moment. maybe someone is interested in doing an extra module for v2/v3.

petrock6:
If someone could recommend a SMBv2/3 library I will start working on it. I cannot find one with an appropriate license. Thanks.

vanhauser-thc:
what is the problem with samba? its GPLv2, hydra is AGPLv3 so there should be no issue at all.

vanhauser-thc:
also there would be http://nicolas.brodu.net/common/programmation/libsmb/archived_site/ but the code is not maintained anymore.

petrock6:
Are you talking about libsmbclient? There's not much documentation for it, and I just ran testbrowse.c to see if it could access a SMB2 Samba share and it didn't work. The examples are from ~2000ish, I'm 99% sure it's deprecated. It did work when I enabled SMB1 though, maybe it could eliminate the issues with multiple tasks on smb1?

Anyway, in the source of Samba, thereis source4/client/client.c which looks promising, but it's a lot of code. For some reason, apt-get install samba on debian doesn't install smbclient (the program -- not the library.) So I can't test it right now. I'll see if I can get the source compiling but it's going to take a while on my dinky raspberry pi.

vanhauser-thc:
if you look here:
https://wiki.samba.org/index.php/Samba3/SMB2
you see that libsmbclient supports SMB v3.11 (and v2 obviously)

apt-get install libsmbclient-dev

petrock6:
Hmmm.. I'll test it again this weekend and get back to you. I'll see if the latest version matches what the Debian repo gave me.

catatonicprime:
@petrock6 how'd the testing go?

rknellx:
[...]

PR #487, opened 2020-01-17
SMB2/3 support

kazkansouh:
Hello,

Recently I came across the issue of needing to pentest against a Windows server that did not have SMB1 enabled. So after looking at #197 , I rolled this simple module for Hydra following the suggested approach in the issue. Although I did look into an alternative approach of using the same samba api as the command line tool `rpcclient`, however, it relied too much on private calls that were not intended to be used outside of internal samba libraries.

I have tested it against a few Windows servers and it appears to work, however I am sure there is room for improvement. Also, I have only tested it compiling/running on Ubuntu 18.04, so there might be issues on other platforms.

In contrast to the original SMB module:
* It is not as precise because it is limited by the exposed api of libsmbclient.
* I did not notice a need to throttle to a single process - maybe as the servers I was testing against were modern (i.e. Windows 2016 and Windows 2019).
* Speed wise, when running as a single process, it is slower, however when running concurrently it is faster - not much surprise.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ hydra-gtk/stamp-h
 pw-inspector
 pw-inspector.exe
 hydra.restore
+*~
--- a/Makefile.am
+++ b/Makefile.am
@@ -21,7 +21,8 @@ SRC = hydra-vnc.c hydra-pcnfs.c hydra-rexec.c hydra-nntp.c hydra-socks5.c \
       hydra-oracle-sid.c hydra-http-proxy.c hydra-http-form.c hydra-irc.c \
       hydra-s7-300.c hydra-redis.c hydra-adam6500.c hydra-rtsp.c \
       hydra-rpcap.c hydra-radmin2.c \
-      hydra-time.c crc32.c d3des.c bfg.c ntlm.c sasl.c hmacmd5.c hydra-mod.c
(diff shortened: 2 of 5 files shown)


vanhauser-thc:
thanks for this.
I think using the samba api (and even private calls - if they are external/exposed/exported its fine, they might change but thats life) would be better, but its your effort so you decide :)
If someone is unhappy with this that person can do it with the samba api ...

vanhauser-thc:
btw - you also need to add it to the xhydra / hydra-gtk list as a module

kazkansouh:
> btw - you also need to add it to the xhydra / hydra-gtk list as a module

Ill take a look into this, thanks!

> I think using the samba api (and even private calls - if they are external/exposed/exported its fine, they might change but thats life) would be better, but its your effort so you decide :)
> If someone is unhappy with this that person can do it with the samba api ...

I agree it would be better, but alas by private calls they have the following issues:
1. Some are exported but the headers are not installed (e.g. `cli_full_connection`). This can be worked around, but alas would increase the maintenance burden on hydra to keep track in case things change (probably the configure script world need to test the installed smb version if there is divergence).
2. Some are not exported and marked as `static`, importantly these are functions that setup the global state of the library, they appear to be called by some call chain that is integrated into common modules `popt` which is used by all the samba command line tools. That said, I am sure there must be a way to hack a call into this, I just stopped looking as it felt like the wrong path to take.

If I find a nice way, Ill look into submitting an `dcerpc` module.

PR #489, opened 2020-01-18
Updated xhydra to support smb2

kazkansouh:
Patch for xhydra (as suggested in #487) to support smb2 module.


vanhauser-thc:
thanks!

PR #492, opened 2020-01-26
Fix -fno-common linker errors

jer-gentoo:
GCC 10 will enable -fno-common by default, which causes linker failures in hydra and xhydra. Fix various declarations and definitions of global variables.

### DIFF
--- a/hydra-asterisk.c
+++ b/hydra-asterisk.c
@@ -9,11 +9,11 @@
 
 extern char *HYDRA_EXIT;
 
-char *buf;
-
 int32_t start_asterisk(int32_t s, char *ip, int32_t port, unsigned char options, char *miscptr, FILE * fp) {
   char *empty = "\"\"";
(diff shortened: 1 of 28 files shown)


vanhauser-thc:
Hi!

thanks a lot for your three PRs, they are highly appreciated!
I too have a list of TODO items from clang-9 and gcc-9 compiler warnings, and its a lot to do ;)
With this patch - I am not sure if moving the *buf into to functions will still work afterwards, one might allocate it, the other free. this will need a more in-depth review.
wouldnt it be easier to just set -fcommon and by this avoid the new gcc-10 default?

jer-gentoo:
Yes, setting `-fcommon` is probably easier than figuring out how the various original declarations, definitions and (re)initialisations would interact.

Mfanakab:
Hello. I'm still a bit new at this but I need to know how ss7 works and can
u guid me through it?

PR #511, opened 2020-04-08
SMB2: improved compatibility when null sessions fail

kazkansouh:
Minor PR that fixes/identifies a case where invalid credentials are not properly identified.


vanhauser-thc:
thank you!

PR #529, opened 2020-05-14
http-proxy: Fix for multiline buffer

kazkansouh:
I was just trying to use `http-proxy` module, but found it was not working as it was assuming the result of `hydra_receive_line` was a single line of text, whereas it actually returns the contents of the receive buffer with null bytes replaced by spaces. Maybe there was some refactoring of the code in the past and `http-proxy` was not updated? I should note as well that I have only tested the basic auth pathways of the code.

Also, added a patch to clean the compile from `sprintf` warnings (at least for `gcc version 9.3.0`).

Finally, added a patch to add http 404 response as a success condition as this is typically from the target server and not the proxy, so its reasonable to assume the authentication was ok in this case. If this is not correct, I can drop this one :)

### DIFF
--- a/hydra-http-proxy.c
+++ b/hydra-http-proxy.c
@@ -7,7 +7,7 @@ char *http_proxy_buf = NULL;
 
 int32_t start_http_proxy(int32_t s, char *ip, int32_t port, unsigned char options, char *miscptr, FILE *fp, char *hostname) {
   char *empty = "";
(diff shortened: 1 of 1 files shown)


Dilatedrose7:
No conflict

vanhauser-thc:
sorry I didnt see this PR. will have a look

vanhauser-thc:
thanks!

Issue #392, opened 2019-03-16
Error: DIGEST-MD5 realm from server could not be extracted


famez:
Hi,

After trying to execute the following command:

**hydra -I -V -s 8443 -l admin -x 4:4:a -m / (ip_addr) https-get**

I get the following error:

**_Error: DIGEST-MD5 realm from server could not be extracted_**

After reviewing the code I found that in the following commit:

https://github.com/vanhauser-thc/thc-hydra/commit/9afbddfa95fbb7bfaad29bbbc5fe926efc66666e#diff-2010e04539ac9f71c48dc3c9244f6e5a

The variable buffer changed from a fixed length array to a pointer.

Line 14:
 ```
 - char *login, *pass, buffer[500], buffer2[500];
  + char *login, *pass, *buffer, buffer2[500];
```

Which makes that in line 65 of file hydra-http.c:

```
#ifdef LIBOPENSSL
  case AUTH_DIGESTMD5:{
      char *pbuffer;

      pbuffer = hydra_strcasestr(http_buf, "WWW-Authenticate: Digest ");
      strncpy(buffer, pbuffer + strlen("WWW-Authenticate: Digest "), sizeof(buffer));
      buffer[sizeof(buffer) - 1] = '\0';

      fooptr = buffer2;
      sasl_digest_md5(fooptr, login, pass, buffer, miscptr, type, webtarget, webport, header);
```

sizeof(buffer) is not 500 anymore but 8. Therefore, only 8 bytes are copied to buffer.

A solution could be:
```
#ifdef LIBOPENSSL
  case AUTH_DIGESTMD5:{
      char *pbuffer;

      pbuffer = hydra_strcasestr(http_buf, "WWW-Authenticate: Digest ");
      strncpy(buffer, pbuffer + strlen("WWW-Authenticate: Digest "), strlen(header) + 499);
      buffer[strlen(header) + 499] = '\0';

      fooptr = buffer2;
      sasl_digest_md5(fooptr, login, pass, buffer, miscptr, type, webtarget, webport, header);
```

**Comment by famez** [comment]
For more details, this is the server`s response:
```
HTTP/1.1 401 Unauthorized
WWW-Authenticate: Digest realm="IE-SR-6GT-LAN", nonce="97499c268ba8de83b87cf8acfce8931c", qop="auth"
Content-Type: text/html
Content-Length: 351
Connection: close
Date: Thu, 15 Feb 2018 11:53:55 GMT
[...]

PR #535, opened 2020-05-21
Update sasl.c

TenGbps:
Some this the realm are long, if is too long is getting a #392


vanhauser-thc:
thank you! I fear that can result in a buffer overflow, so I will add some length limits.

vanhauser-thc:
I pushed a commit, can you please test that it still works for you?

PR #538, opened 2020-06-08
Updating hydra support for freedrdp 3

animetauren:
The newest version of freerdp is v.3, updating Hydra to support this new module. 

Have tested this on my Ubuntu 18.04 LTS using latest build from freerdp and thc-hydra at time of writing.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -78,6 +78,7 @@ install:	strip
 	-cp -f *.csv $(DESTDIR)$(PREFIX)$(DATADIR)
 	-mkdir -p $(DESTDIR)$(PREFIX)$(MANDIR)
 	-cp -f hydra.1 xhydra.1 pw-inspector.1 $(DESTDIR)$(PREFIX)$(MANDIR)
+      -ldconfig
 
 clean:
 	rm -rf xhydra pw-inspector hydra *.o core *.core *.stackdump *~ Makefile.in Makefile dev_rfc hydra.restore arm/*.ipk arm/ipkg/usr/bin/* hydra-gtk/src/*.o hydra-gtk/src/xhydra hydra-gtk/stamp-h hydra-gtk/config.status hydra-gtk/errors hydra-gtk/config.log hydra-gtk/src/.deps hydra-gtk/src/Makefile hydra-gtk/Makefile
--- a/configure
+++ b/configure
@@ -70,8 +70,8 @@ NSL_PATH=""
 SOCKET_PATH=""
 MANDIR=""
 XHYDRA_SUPPORT=""
-FREERDP2_PATH=""
-WINPR2_PATH=""
+FREERDP3_PATH=""
+WINPR3_PATH=""
 SMBC_PATH=""
 SMBC_IPATH=""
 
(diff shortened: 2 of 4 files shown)


vanhauser-thc:
so everyone who runs a Linux that does not have freerdp3 yet cant use rdp anymore?
do you think that is a good solution?

animetauren:
> so everyone who runs a Linux that does not have freerdp3 yet cant use rdp anymore?

Not sure what you are getting at, just if you are running thc-hydra and pull the latest freerdp repo and build from source, you will get an issue. I did, found the issue and just posted a solution that meant adding support for the new version. 

>

vanhauser-thc:
a solution that breaks rdp for most users is not a solution but a bug :)
the correct fix is changing the configure script to test if librdp3 is present and if not if librdp2 is present. not breaking it for librdp2 users.

animetauren:
Let me go that, and re-submit a proper PR, good call absolutely right

hydra-rdp.c
```diff
@@ -10,7 +10,7 @@
 #include "hydra-mod.h"
 
 extern char *HYDRA_EXIT;
-#ifndef LIBFREERDP2
+#if !defined(LIBFREERDP2) || (LIBFREERDP3)
```
vanhauser-thc, line 13: if the library versions make no different is v2 or v3 I would just name it LIBFREERDP

configure
```diff
@@ -1599,11 +1672,11 @@ fi
...
 fi
-if [ -n "$FREERDP2_PATH" ]; then
-    XLIBS="$XLIBS -lfreerdp2"
+if [ -n "$FREERDP3_PATH" ]; then
+    XLIBS="$XLIBS -lfreerdp3"
 fi
-if [ -n "$WINPR2_PATH" ]; then
-    XLIBS="$XLIBS -lwinpr2"
+if [ -n "$WINPR3_PATH" ]; then
+    XLIBS="$XLIBS -lwinpr3"
```
vanhauser-thc, line 1679: here you have not reestablished the previous librdp2 for XLIBS, only librdp3

hydra-rdp.c
animetauren: Agreed, just fixed this

configure
animetauren: Just fixed the XLIBS and XDEFINES paths

animetauren:
@vanhauser-thc this should be it, tested this on my machine, compiles and runs as it should and fixed the ldconfig issue by adding the instructions to the makefile. Let me know your thoughts.

Makefile.am
```diff
@@ -78,6 +78,7 @@ install:	strip
 	-cp -f *.csv $(DESTDIR)$(PREFIX)$(DATADIR)
 	-mkdir -p $(DESTDIR)$(PREFIX)$(MANDIR)
 	-cp -f hydra.1 xhydra.1 pw-inspector.1 $(DESTDIR)$(PREFIX)$(MANDIR)
+      -ldconfig
```
vanhauser-thc, line 81: why that? that would only be needed if a new shared library would be installed (which hydra does not have).

configure
```diff
@@ -1518,9 +1601,12 @@ fi
 if [ -n "$MONGODB_IPATH" ]; then
     XIPATHS="$XIPATHS -I$MONGODB_IPATH -I$BSON_IPATH"
 fi
-if [ -n "$FREERDP2_IPATH" ]; then
+if [ -n "$FREERDP3_IPATH" ]; then
```
vanhauser-thc, line 1604: this one is wrong, you moved the check to RDP3, but then copy RDP2_IPATH. as the real RDP3 one is below this has to be $FREERDP2_IPATH :)

vanhauser-thc:
besides two minor things I just commented on it looks good.
One issue that could result is that if librdp2 and 3 are both installed then both will tried to be linked and that would fail. and you always check for both being present.
you could also check for 3 first and then skip 2 if it was found, that would save the trouble.

Makefile.am
animetauren: This is needed for "libfreerdp3.so.3" which is loaded as a shared library, not sure why this is the case for v3 and not for v2. This is the error I get when ldconfig is not done:

` /usr/local/bin/hydra: error while loading shared libraries: libfreerdp3.so.3: cannot open shared object file: No such file or directory`

configure
vanhauser-thc: this bug is still present

vanhauser-thc:
sorry for the long delay, I have so much to do ...

there is one bug left, just commented on it.
also if both libfreerdp2 and libfreerdp3 are present the script will try to link both. I would recommend to prefer rdp3 over rdp2.

configure
animetauren: Working on this today and will update when finished

animetauren:
@vanhauser-thc This should be it, fixed the bug you mentioned, added logic to handle version of freerdp properly and just updated bash conditionals.

vanhauser-thc:
looks good, thanks!

vanhauser-thc:
I had to perform a few fixes:
 * my `[` does not understand &&/|| and failed, reversed that change
 * the logic to test for freerdp2 was inverse hence it was only checked for if freerdp3 was found
 * the ldconfig was not indented which makes gnu make fail. also does belong to the library install not hydra

but now everything compiles cleanly, thanks!

Issue #530, opened 2020-05-14
No help for https-post-form


Jab2870:
Hi,

From various blogs and looking at other issues, I can see that the https-post-form method should be used with the same options as http-post-form. That is fine and works although the -U option doesn't show help.

```
$ hydra -U http-post-form
Hydra v9.0 (c) 2019 by van Hauser/THC - Please do not use in military or secret service organizations, or for illegal purposes.

Hydra (https://github.com/vanhauser-thc/thc-hydra) starting at 2020-05-14 14:09:37

Help for module http-post-form:
============================================================================
Module http-post-form requires the page and the parameters for the web form.

By default this module is configured to follow a maximum of 5 redirections in
a row. It always gathers a new cookie from the same URL without variables
The parameters take three ":" separated values, plus optional values.
(Note: if you need a colon in the option string as value, escape it with "\:", but do not escape a "\" with "\\".)

Syntax:   <url>:<form parameters>:<condition string>[:<optional>[:<optional>]
First is the page on the server to GET or POST to (URL).
Second is the POST/GET variables (taken from either the browser, proxy, etc.
 with url-encoded (resp. base64-encoded) usernames and passwords being replaced in the
 "^USER^" (resp. "^USER64^") and "^PASS^" (resp. "^PASS64^") placeholders (FORM PARAMETERS)
Third is the string that it checks for an *invalid* login (by default)
 Invalid condition login check can be preceded by "F=", successful condition
 login check must be preceded by "S=".
 This is where most people get it wrong. You have to check the webapp what a
 failed string looks like and put it in this parameter!
The following parameters are optional:
 (c|C)=/page/uri     to define a different page to gather initial cookies from
 (h|H)=My-Hdr\: foo   to send a user defined HTTP header with each request
                 ^USER[64]^ and ^PASS[64]^ can also be put into these headers!
                 Note: 'h' will add the user-defined header at the end
                 regardless it's already being sent by Hydra or not.
                 'H' will replace the value of that header if it exists, by the
                 one supplied by the user, or add the header at the end
Note that if you are going to put colons (:) in your headers you should escape them with a backslash (\).
 All colons that are not option separators should be escaped (see the examples above and below).
 You can specify a header without escaping the colons, but that way you will not be able to put colons
 in the header value itself, as they will be interpreted by hydra as option separators.

Examples:
 "/login.php:user=^USER^&pass=^PASS^:incorrect"
[...]

PR #542, opened 2020-06-26
Usage: Fix help for https-post-form and https-get-form

Jab2870:
Fixes issue #530

### DIFF
--- a/hydra.c
+++ b/hydra.c
@@ -619,10 +619,8 @@ void module_usage() {
          "%s:\n================================================================"
          "============\n",
          hydra_options.service);
-  if (strcmp(hydra_options.service, "https-post-form") == 0)
-    strcpy(hydra_options.service, "http-post-form");
-  else if (strcmp(hydra_options.service, "https-get-form") == 0)
-    strcpy(hydra_options.service, "http-get-form");
+  if (strncmp(hydra_options.service, "https-", 6) == 0 )
+    memmove(hydra_options.service + 4, hydra_options.service + 5, strlen(hydra_options.service) - 4);
   for (i = 0; i < sizeof(services) / sizeof(services[0]); i++) {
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
thank you!
I rather spend my team teaching you how to do a good fix rather than doing it myself (doing it myself would be 10x faster ;)

your solution is OK, it is not pretty though as you are overwriting the name of the service. but as after the help output hydra exits that is not an issue.

if you implement it like this though - maybe you should also handle other protocols there that share the same issue?
e.g. https-post/get/head ...
and then you could make your solutions easier -> e.g. if it starts with https- then copy the string therafter directly to http- (this can be done with the memove function on the same variable ... memmove(service + 4, service + 5, strlen(service) - 5)

Jab2870:
> I rather spend my team teaching you how to do a good fix rather than doing it myself (doing it myself would be 10x faster ;)

Thank you very much

> your solution is OK, it is not pretty though as you are overwriting the name of the service. but as after the help output hydra exits that is not an issue.

I did wonder if overwriting the name would be an issue, but I saw that the same was done here:

https://github.com/vanhauser-thc/thc-hydra/blob/master/hydra.c#L3113-L3126

I didn't worry about `hydra_options.ssl` because I saw that it ran exit and wasn't used beforehand.

I have made the change you suggested so now any service starting with `https-` will be changed to `http-` for the help section

vanhauser-thc:
did you test it?
because I think I made a mistake with what I wrote. length of menmove should be strlen() - -4 + 1 as otherwise the terminating null byte is not copies too and then the service name has the last character as a double.

Jab2870:
Yes, I tested it and it appears to work.

I did change your code slightly because the last character was duplicated:

```
    memmove(hydra_options.service + 4, hydra_options.service + 5, strlen(hydra_options.service) - 4);
```

PR #562, opened 2020-09-09
Bruteforce module

e2002e:
Hi, you pulled the part about 'rain' algorithm some time ago.
It was supposed to delinearize the bruteforce, but does not do so, instead it will create some variations for the non yet iterating
part of the word.

I realized that and managed to have the index of the iteration dispatched across the word; that combined with a rotation has a good effect.

I have thoroughly  tested the same exact code that I implemented in my fork of john the ripper. No errors.
Also launched a small brute on my ssh to ensure it was correct and it is.

Regards.


e2002e:
I have done the last fix, you can pull it and the rain option is effective, or just remove parts of the code that we included beforehand.
Cheers.

vanhauser-thc:
hi!
please do a make clean and resubmit, the Makefile is overwritten in this commit. I make the same mistake quite often too :)
also - please take care of the conflicts.

e2002e:
Commited the untouched Makefile, also resolved the conflicts.
The statement that starts a 1290 in hydra_http_form.c is commented and was replaced but there was an issue. Are you sure this module is working ?

vanhauser-thc:
thank you!

```
  if(rainy)  {
  	int mpldisp = bf_options.current/2+3;
	int mplmod2 = bf_options.current % 2;
	int strafeIndex;		
	for(i=0; i<bf_options.current; ++i) {
		if(mplmod2) strafeIndex  = (bf_options.strafe+i)%bf_options.current;
		else strafeIndex = (i+mpldisp)%bf_options.current;

		bf_options.ptr[i] = bf_options.crs[(bf_options.state[strafeIndex] + bf_options.rotate) % bf_options.crs_len];
		bf_options.rotate += i+1;
		bf_options.strafe += 3;
	}
	#define accu(i) \
	do { \
		int j; \
		for(j=1; j<=i; ++j) k += j; \
	} while(0)

	int k = 0;
	accu(bf_options.current);
	bf_options.rotate -= k-4;
  }
```
this is the main change. can you explain what happens here and how it changes the behaviour in contrast to the previous code?

e2002e:
Hello, I have close/reopened (don't know if that is necessary) because there is a last commit that fixes a bug.
To explain the method: there are two complementary things, as you can see there is a variable 'rotate' and a variable 'strafe'.
The strafe is used with the state this way: "(bf_options.state[(bf_options.strafe[loop]/(bf_options.current/4)+i". And is modified according to the length of the word at each characters: bf_options.strafe += 2 * (2-bf_options.current%2);
This does make the index of the state slide along the words as, you can see in [this video](https://youtu.be/Vqf8Bjyl1Dk).
The rotation is added above this to make it more efficient, but as I said in the first PR message, I realized that this is only a rotation, not some kind of magic...

vanhauser-thc:
there is not need for close and reopen.

e2002e:
I am working on the rotation, which produces duplicates in some corner cases, I'll be reopening this when fixed.
bis nachste mäl.

PR #569, opened 2020-09-29
Rain mode

e2002e:
Ok vanhauser, I had told you about the rain mode being ineffective, so I searched and tried more and found the correct solution.
You can see for yourself that the module is faultless and what the generation does look like by compiling my testbfg.c with -lm.
Right now I'm trying to convince the maintainers of john the ripper to merge my fork.
I hope that you will find some understanding of the code, and agree to consider it a valid mangling solution.
Regards.
[test.zip](https://github.com/vanhauser-thc/thc-hydra/files/5301333/test.zip)


hydra.c
```diff
@@ -591,7 +591,8 @@ void help_bfg() {
          "             'A' for uppercase letters, '1' for numbers, and for all "
          "others,\n"
          "             just add their real representation.\n"
-         "  -y         disable the use of the above letters as placeholders\n\n"
+         "  -y         disable the use of the above letters as placeholders\n"
+         "  -r         use a method to delinearize the bruteforce.\n\n"
```
vanhauser-thc, line 595: that description basically tells the user nothing :) why should he/she use the option? and what would be the result?

e2002e: It's true that until you see an output it's hard to guess.
Maybe the idea of an increasing value used to modify the state and cut with a modulo, 
I'd have to explain that there is a slide in the sequence of increasing and decreasing the value.
The code is rain = i+3 in the loop on the length of the word,
When subtracting; for even words starting the loop at 1 and using i+1 can be explained as "subtract the same value that we added, minus (length-1) * 2; 
for odd words the loop for subtraction starts at 2 and it's end is inclusive, we are still using i+1 but i'm having a hard time seeing what it does.
I made a last fix that i'm pushing now/

vanhauser-thc: you have to change the description to something meaningful for the user. the text there now could also be "foo bar foo bar" :)

vanhauser-thc:
I am gone for 1 1/2 weeks, do some more testing so I can merge when I am back

e2002e:
Ok I will, 
I just figured out the subtraction for odd numbers, it is equivalent to 'sum that we added in the loop' - (length-1)

e2002e:
It was worth doing more testings, I found what falls right with a deviation on the index of the state and another one on the value of the state, I does 'strafe' the index being treated and 'rotate' the selection of the character in the set, hence the variable names.
This will not work on works shorter than 3, so I added an `if` that will take care of that.

e2002e:
I'm good van hauser, please don't trust the file I send you to check the algorithm.
I had to change it, I'm still cutting a strafe called 'rain' with a modulo of the length of the character set. But this strafe is divided by two for each character in the word but the first one, which isn't in the loop on bf_options.current. I don't know why it doesn't work otherwise. When a word is done we set it to the current word's counter value + 1.
The state is worked on in the other way, to temper with the 1th character being static.
Please refer to my software at https://github.com/e2002e/zhou if you want to see it run or check the word's count by yourself.
Gut Nacht

vanhauser-thc:
:)
do so!

e2002e:
Ich habe das solution fur die problem of false shuffling.
Tested with my fork of johntheripper and rainycrack.
./rc 10 10 -s abcdefghij 
from the output, picking a pattern and ./rc 10 10 -s abcdefghij |grep pattern
Doesn't repeat.
Word counts are good too.
I leave it to you.

e2002e:
brute

vanhauser-thc:
@e2002e you must first resolve the merge conflict before I can look at it and test it.

e2002e:
@vanhauser-thc done !
As I remember the post form was broken something like one year ago so I had to change the code in my fork.

vanhauser-thc:
sorry that it took so long. so much to do ...

e2002e:
Thanks for your trust vh.
I'm planning on being more with nature than with my laptop now. Working on
the PC takes abstraction from it at one point.
Auf viedersehen.

PR #579, opened 2020-11-12
http-form: ipv6 fix 

kazkansouh:
Hello,

I noticed that when using the http-form module with an ipv6 address that it was failing to work. After investigation, it appears the `Host` header was not correctly formatted, resulting in the following:

Screenshot_20201111_142823

Essentially, it should include the square brackets around the address. I noticed another issue on another project that had the same problem and provides links to the RFC's https://github.com/dotnet/runtime/issues/25661#issuecomment-377023158

While tweaking the `Host` header, I noticed the following *if* statement which complicated the code and as far as I can tell guarded dead code (as it expects the `miscptr` to contain `://`) and the only effect of the statement was to set `webtarget = null` due to `strstr`.
https://github.com/vanhauser-thc/thc-hydra/blob/c2260d2c5d55a40e5ed51d9f995b31e425a095d1/hydra-http-form.c#L1247

I tried different things to execute this code guarded by the above *if*, with the most obvious being something of the following form:
```plaintext
$ hydra -l a -p a 'http-form-post://ptsv2.com/http://ptsv2.com/t/5sysq-1605183971/post:user=^USER^&password=^PASS^:Sorry'
```
But this just confused some code in `hydra.c`:
```plaintext
[ERROR] the variables argument needs at least the strings ^USER^, ^PASS^, ^USER64^ or ^PASS64^: //ptsv2.com/t/5sysq-1605183951/post
```
In the end, I just removed it as it simplified the implementation. If I'm mistaken and this *if* statement is useful, let me know and Ill update this PR :wink:

Finally, the second commit contains two minor changes to fix warnings flagged up by the compiler.

### DIFF
--- a/hydra-http-form.c
+++ b/hydra-http-form.c
@@ -1039,7 +1039,7 @@ int32_t start_http_form(int32_t s, char *ip, int32_t port, unsigned char options
       }
 
       if (strrchr(url, ':') == NULL && port != 80) {
-        sprintf(str2, "%s:%d", str2, port);
+        sprintf(str2, "%.2040s:%d", str2, port);
       }
 
       if (verbose)
--- a/hydra-http-form.c
+++ b/hydra-http-form.c
@@ -78,7 +78,7 @@ int32_t auth_flag = 0;
 
 char cookie[4096] = "", cmiscptr[1024];
 
-int32_t webport, freemischttpform = 0;
+int32_t webport;
 char bufferurl[6096 + 24], cookieurl[6096 + 24] = "", userheader[6096 + 24] = "", *url, *variables, *optional1;
 
 #define MAX_REDIRECT 8
@@ -1133,9 +1133,6 @@ void service_http_form(char *ip, int32_t sp, unsigned char options, char *miscpt
   while (1) {
     if (run == 2) {
       if (memcmp(hydra_get_next_pair(), &HYDRA_EXIT, sizeof(HYDRA_EXIT)) == 0) {
-        if (freemischttpform)
-          free(miscptr);
-        freemischttpform = 0;
         hydra_child_exit(1);
       }
     }
@@ -1157,9 +1154,6 @@ void service_http_form(char *ip, int32_t sp, unsigned char options, char *miscpt
       }
       if (sock < 0) {
         hydra_report(stderr, "[ERROR] Child with pid %d terminating, cannot connect\n", (int32_t)getpid());
-        if (freemischttpform)
-          free(miscptr);
-        freemischttpform = 0;
         hydra_child_exit(1);
       }
       next_run = 2;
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
Hi, you made lots of other changes in the code ... why is that?

kazkansouh:
Ok, no problem let me try and explain.

So the core issue was that I needed to use `malloc` to allocate a new `webtarget` large enough for the `[` and `]` (also I took the opportunity to add the port to the `webtarget`, if non-standard, as [it should be][rfc] there).

The *if* statement below complicated things:
https://github.com/vanhauser-thc/thc-hydra/blob/c2260d2c5d55a40e5ed51d9f995b31e425a095d1/hydra-http-form.c#L1247-L1267

This is because:
1. The `initialize` function is called a couple times, where first time `webtarget == NULL` and second time its not, however the *if* statement will essentially set it to `NULL` in the case `miscptr` does not contain `://` due to the use of `strstr`. So with this *if* statement present, on subsequent calls, the `webtarget` would need to be free'ed before its null'ed.
2. Reading the implementation of the *if* statement, it looks like its intended that the `miscptr` can be used to also specify the `webtarget`, and hence the host header e.g. setting a module option of `http://abc/login:user=^USER^......` instead of just `/login:user=^USER^......`. However, I was unable to get this functionality to work - and its not documented in the module help. So I concluded it is old functionality that is no longer supported.
3. Scanning `miscptr` for `://` is a potential issue in-case a parameter/cookie or other item contains this string ... not sure if this is a real concern, but its worth mentioning.

Thus, for these reasons, I decided to strip out the above *if* statement and replace it with code that sets `webtarget` to `NULL`, and thus all the references to `freemischttpform` were not required any more - simplifying the code a little bit :tada: That covers most of the changes, except for the change to `hydra.c`:

This last change (which is part of the second commit) was because `gcc` flagged it as a warning and so I stripped out as it was easy to do:
https://github.com/vanhauser-thc/thc-hydra/blob/c2260d2c5d55a40e5ed51d9f995b31e425a095d1/hydra.c#L613-L616 

Basically, the second parameter to `printf` will always be a null pointer :facepalm: However, its never executed as this case is already covered with the following:
https://github.com/vanhauser-thc/thc-hydra/blob/c2260d2c5d55a40e5ed51d9f995b31e425a095d1/hydra.c#L2699-L2706
So it should be safe to remove without any noticeable effect.

I hope this helps explain the changes. If I made a mistake, just let me know and Ill refactor the patch.

[rfc]: https://tools.ietf.org/html/rfc7230#section-5.4

vanhauser-thc:
hi!
sorry for taking so long, I am flooded with work ...
I accepted the PR, can you please send a 2nd PR for hydra-http.c ? Same IPv6 issue should be in there too.

kazkansouh:
> hi!
> sorry for taking so long, I am flooded with work ...
> I accepted the PR, can you please send a 2nd PR for hydra-http.c ? Same IPv6 issue should be in there too.

Hello, no problem. Hopefully will submit it in a few days :)

PR #582, opened 2020-12-04
cleanup

e2002e:
Hallo, I have left a useless procedure in the code, this commit is removing it.
No changes otherwise.

### DIFF
--- a/bfg.c
+++ b/bfg.c
@@ -202,14 +202,6 @@ uint64_t bf_get_pcount() {
   return foo;
 }
 
-
-int accu(int x) {
-    int a = 0, b;
-    for(b=1; b<x; ++b)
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
it is not a cleanup it is either a fix or you introduced a bug as the behavior changes. what is it?

e2002e:
No actually that is was I wanted to say there should be no behaviour's
change.
The counter used to set the 'rain' variable, with which we play inside the
iteration on word, is increased by one each word, instead of being
increased inside that loop to then be reduced to match an increase of 1.
This was pointless as we are not using the counter inside the iteration so
if there was a change in behaviour this means that the 'accu' function
wasn't well writen.

PR #586, opened 2020-12-26
http-get: ipv6 fix

kazkansouh:
As requested in https://github.com/vanhauser-thc/thc-hydra/pull/579#issuecomment-735768540 this PR aligns the http-get module with the patch for http-form module in #579.


vanhauser-thc:
thank you!

PR #610, opened 2021-03-03
STDIN (save yourself from the rain !)

e2002e:
Hi vh, lattely I was having regrets about my rain mode merged with bfg.
Indeed this ALGORITHM is NOT SUITABLE for online password cracking. Even the guys at openwall don't want it because it doesn't crack keys "fast enough" localy, here you are dealing with remote crack without hashes ...so DO NOT KEEP it please.
In this pull request I have utterly removed the 'rain' algorithm, setting the bfg back to it's original state.
My first idea more than 5 years ago was simply to let hydra manage std input, so I come to you with this done (that's really cool you can pipe john's incremental mode through hydra). This comes with a flaw because hydra needs to know how much password it is going to test before it starts, so you have to pass the numbers of lines wanted on the cmdline.
Tell me what you think about this.

### DIFF
--- a/bfg.c
+++ b/bfg.c
@@ -174,8 +174,6 @@ int32_t bf_init(char *arg) {
 	
   bf_options.crs_len = crs_len;
   bf_options.current = bf_options.from;
-  bf_options.rain = 0;
-  bf_options.gcounter = 0;
 
   memset((char *) bf_options.state, 0, sizeof(bf_options.state));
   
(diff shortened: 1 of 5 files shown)


vanhauser-thc:
I see no value in reading passwords from stdin. never have, never will. you can just write that input to a file and pass that to hydra.
also you remove an important change from configure.

e2002e:
Then, i'll just make a pull request with the algo removed. You can't keep
that vh.

Le jeu. 11 mars 2021 à 13:55, van Hauser ***@***.***> a
écrit :

PR #638, opened 2021-04-26
Fix logic bug

chris-pcguy:
I can provide another patch version, which swaps the operands instead.

Signed-off-by: Christian Inci <chris.gh@broke-the-inter.net>


vanhauser-thc:
thank you!

PR #648, opened 2021-05-12
Changed index() to strchr()

sanmacorz:
Changed index() to strchr() because the index() function seems to be deprecated in some POSIX systems. Don't know if this affects any other part of the program.


vanhauser-thc:
Thank you! I am away and will review it on Monday

vanhauser-thc:
looks good, thank you!

PR #651, opened 2021-05-19
fix: skip user bug

wj0s3ph:
username can potentially be identical to the beginning of login_ptr


vanhauser-thc:
I do not see how this is a fix because the exact same thing is happening :)
but strcmp makes more sense.

PR #669, opened 2021-07-02
Optimize burst http service and telnet service

boynicholas:
Some http authentication servers will return a non-401 status code when the password is attempted more than once, which leads to false positives, Of course the vast majority of me doesn't know what the successful login reply content, so I can't tell by the content

The addition of parameters makes it possible to determine success or failure based on the reply status code:

`(c|C)=check for status code in the HTTP reply. If the reply status code
            matches it, it is judged as successful`

---
Some telnet servers will have a % sign in the prefix after a failed login, which also leads to false success reports, which are dealt with it

### DIFF
--- a/hydra-http.c
+++ b/hydra-http.c
@@ -9,6 +9,7 @@ char *http_buf = NULL;
 #define END_CONDITION_MAX_LEN 100
 static char end_condition[END_CONDITION_MAX_LEN];
 int end_condition_type = -1;
+int match_status_code = 0;
 
 int32_t webport;
 int32_t http_auth_mechanism = AUTH_UNASSIGNED;
@@ -248,6 +249,25 @@ int32_t start_http(int32_t s, char *ip, int32_t port, unsigned char options, cha
   ptr = ((char *)strchr(http_buf, ' '));
   if (ptr != NULL)
     ptr++;
+  // check status
+  if (match_status_code != 0) {
+    char *status_code;
+    memset(status_code, 0, 3);
+
+    for (int i = 0; i < 3; i++) {
+      *(status_code + i) = *(ptr + i);
+    }
+
+    if (match_status_code == atoi(status_code)) {
+      if (end_condition == -1) {
+        if (debug)
+          hydra_report(stderr, "Status code condition %d match.\n", status_code);
+        hydra_report_found_host(port, ip, "www", fp);
+        hydra_completed_pair_found();
+      }
+    }
+  }
+
   if (ptr != NULL && (*ptr == '2' || *ptr == '3' || strncmp(ptr, "403", 3) == 0 || strncmp(ptr, "404", 3) == 0)) {
 #ifdef HAVE_PCRE
     if (end_condition_type >= 0 && hydra_string_match(http_buf, end_condition) != end_condition_type) {
@@ -426,37 +446,67 @@ int32_t service_http_init(char *ip, int32_t sp, unsigned char options, char *mis
   //   -1  error, hydra will exit, so print a good error message here
 
   /*POU CODE */
-  char *start = strstr(miscptr, "F=");
-  if (start == NULL)
-    start = strstr(miscptr, "S=");
 
-  if (start != NULL) {
-    if (start[0] == 'F')
-      end_condition_type = 0;
-    else
-      end_condition_type = 1;
+  char *delim = ":";
+  strtok(miscptr, delim);
+  char *p;
 
-    int condition_len = strlen(start);
-    memset(end_condition, 0, END_CONDITION_MAX_LEN);
-    if (condition_len >= END_CONDITION_MAX_LEN) {
-      hydra_report(stderr, "Condition string cannot be bigger than %u.", END_CONDITION_MAX_LEN);
-      return -1;
+  while ((p = strtok(NULL, delim))) {
+    if (strstr(p, "F=") != NULL || strstr(p, "S=") != NULL) {
+      char *match_text_start = strstr(p, "F=");
+      if (match_text_start == NULL)
+        match_text_start = strstr(p, "S=");
+
+      if (match_text_start != NULL) {
+        if (match_text_start[0] == 'F')
+          end_condition_type = 0;
+        else
+          end_condition_type = 1;
+
+        int condition_len = strlen(match_text_start);
+        memset(end_condition, 0, END_CONDITION_MAX_LEN);
+        if (condition_len >= END_CONDITION_MAX_LEN) {
+          hydra_report(stderr, "Condition string cannot be bigger than %u.", END_CONDITION_MAX_LEN);
+          return -1;
+        }
+        // copy condition witout starting string (F= or S=  2char)
+        strncpy(end_condition, match_text_start + 2, condition_len - 2);
+        if (debug)
+          hydra_report(stderr, "End condition is %s, mod is %d\n", end_condition, end_condition_type);
+
+        if (*(match_text_start - 1) == ' ')
+          match_text_start--;
+        memset(match_text_start, '\0', condition_len);
+        if (debug)
+          hydra_report(stderr, "Modificated options:%s\n", miscptr);
+      } else {
+        if (debug)
+          hydra_report(stderr, "Condition not found\n");
+      }
     }
-    // copy condition witout starting string (F= or S=  2char)
-    strncpy(end_condition, start + 2, condition_len - 2);
-    if (debug)
-      hydra_report(stderr, "End condition is %s, mod is %d\n", end_condition, end_condition_type);
 
-    if (*(start - 1) == ' ')
-      start--;
-    memset(start, '\0', condition_len);
-    if (debug)
-      hydra_report(stderr, "Modificated options:%s\n", miscptr);
-  } else {
-    if (debug)
-      hydra_report(stderr, "Condition not found\n");
+    if (strstr(p, "c=") != NULL || strstr(p, 'C=') != NULL) {
+      if (strlen(p) != 2 + 3) {
+        hydra_report(stderr, "Invalid status code, eg: 200.");
+        return -1;
+      }
+
+      char *status_code;
+      memset(status_code, 0, 3);
+
+      strncpy(status_code, p + 2, 3);
+
+      for (int i = 0; i < strlen(status_code); i++) {
+        if (!isdigit(*(status_code + i))) {
+          hydra_report(stderr, "Invalid status code, eg: 200");
+          return -1;
+        }
+      }
+
+      match_status_code = atoi(status_code);
+      free(status_code);
+    }
   }
-
   return 0;
 }
 
@@ -467,6 +517,8 @@ void usage_http(const char *service) {
          "NTLM or MD5\n"
          " (h|H)=My-Hdr\\: foo   to send a user defined HTTP header with each "
          "request\n"
+         " (c|C)=check for status code in the HTTP reply. If the reply status "
+         " code matches it, it is judged as successful \n"
          " (F|S)=check for text in the HTTP reply. S= means if this text is "
          "found, a\n"
          "       valid account has been found, F= means if this string is "
--- a/hydra-http.c
+++ b/hydra-http.c
@@ -504,7 +504,10 @@ int32_t service_http_init(char *ip, int32_t sp, unsigned char options, char *mis
       }
 
       match_status_code = atoi(status_code);
-      free(status_code);
+      
+      if (debug) {
+        hydra_report(stdout, "Find match reply status code: %d", match_status_code);
+      }
     }
   }
   return 0;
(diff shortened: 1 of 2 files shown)


hydra-http.c
```diff
@@ -248,6 +249,25 @@ int32_t start_http(int32_t s, char *ip, int32_t port, unsigned char options, cha
   ptr = ((char *)strchr(http_buf, ' '));
   if (ptr != NULL)
     ptr++;
+  // check status
+  if (match_status_code != 0) {
+    char *status_code;
+    memset(status_code, 0, 3);
```
vanhauser-thc, line 255: status_code is an unassigned pointer and you do a memset on it - this crashes

hydra-http.c
```diff
@@ -248,6 +249,25 @@ int32_t start_http(int32_t s, char *ip, int32_t port, unsigned char options, cha
...
+
+    for (int i = 0; i < 3; i++) {
+      *(status_code + i) = *(ptr + i);
+    }
+
```
vanhauser-thc, line 260: why this hassle and not just calling atoi(ptr)?

hydra-http.c
```diff
@@ -426,37 +446,70 @@ int32_t service_http_init(char *ip, int32_t sp, unsigned char options, char *mis
...
+  strtok(miscptr, delim);
+  char *p;
+
+  while ((p = strtok(NULL, delim))) {
+    if (strstr(p, "F=") != NULL || strstr(p, "S=") != NULL) {
+      char *match_text_start = strstr(p, "F=");
+      if (match_text_start == NULL)
+        match_text_start = strstr(p, "S=");
+
+      if (match_text_start != NULL) {
```
vanhauser-thc, line 460: this is impossible to be NULL as otherwise you would not have passed the above `if (strstr(...`

hydra-http.c
```diff
@@ -426,37 +446,70 @@ int32_t service_http_init(char *ip, int32_t sp, unsigned char options, char *mis
...
+          end_condition_type = 1;
+
+        int condition_len = strlen(match_text_start);
+        memset(end_condition, 0, END_CONDITION_MAX_LEN);
+        if (condition_len >= END_CONDITION_MAX_LEN) {
+          hydra_report(stderr, "Condition string cannot be bigger than %u.", END_CONDITION_MAX_LEN);
+          return -1;
+        }
+        // copy condition witout starting string (F= or S=  2char)
+        strncpy(end_condition, match_text_start + 2, condition_len - 2);
```
vanhauser-thc, line 473: no need to memset and strncpy if you already have checked if condition_len >= END_CONDITION_LEN.  you can directly strcpy() instead.

hydra-http.c
```diff
@@ -426,37 +446,70 @@ int32_t service_http_init(char *ip, int32_t sp, unsigned char options, char *mis
...
+        if (debug)
+          hydra_report(stderr, "End condition is %s, mod is %d\n", end_condition, end_condition_type);
+
+        if (*(match_text_start - 1) == ' ')
```
vanhauser-thc, line 477: if F=/S= is the first parameter then this is an out of bounds read. and why this check?

hydra-http.c
```diff
@@ -426,37 +446,70 @@ int32_t service_http_init(char *ip, int32_t sp, unsigned char options, char *mis
...
+          match_text_start--;
+        memset(match_text_start, '\0', condition_len);
```
vanhauser-thc, line 479: why the memset?

hydra-http.c
```diff
@@ -426,37 +446,70 @@ int32_t service_http_init(char *ip, int32_t sp, unsigned char options, char *mis
...
-    if (*(start - 1) == ' ')
-      start--;
-    memset(start, '\0', condition_len);
-    if (debug)
-      hydra_report(stderr, "Modificated options:%s\n", miscptr);
-  } else {
-    if (debug)
-      hydra_report(stderr, "Condition not found\n");
-  }
+    if (strstr(p, "c=") != NULL || strstr(p, 'C=') != NULL) {
```
vanhauser-thc, line 488: C is already used for setting cookies

hydra-http.c
```diff
@@ -467,6 +520,8 @@ void usage_http(const char *service) {
          "NTLM or MD5\n"
          " (h|H)=My-Hdr\\: foo   to send a user defined HTTP header with each "
          "request\n"
+         " (c|C)=check for status code in the HTTP reply. If the reply status code\n"
```
vanhauser-thc, line 523: again, this collides with http-form module:
```
 (c|C)=/page/uri     to define a different page to gather initial cookies from
```

vanhauser-thc: maybe R for result code

hydra-http.c
```diff
@@ -426,37 +446,70 @@ int32_t service_http_init(char *ip, int32_t sp, unsigned char options, char *mis
...
-    if (condition_len >= END_CONDITION_MAX_LEN) {
-      hydra_report(stderr, "Condition string cannot be bigger than %u.", END_CONDITION_MAX_LEN);
-      return -1;
+
+  char *delim = ":";
+  strtok(miscptr, delim);
+  char *p;
+
+  while ((p = strtok(NULL, delim))) {
+    if (strstr(p, "F=") != NULL || strstr(p, "S=") != NULL) {
```
vanhauser-thc, line 455: you are changing the behaviour here.
the S=/F= is optional, is no (char)=  is given it defaults to F=
also note for F=/S=:`Note: this must be the last option supplied.`
But I would not mind to change this, that always F=/S= has to be given and that it doesnt need to be the last parameter. But then a) the documentation has to be updated (help output) and b) the http-form module updated as well.

vanhauser-thc:
@boynicholas thank you this looks useful. It needs some more work though.

boynicholas:
Sorry, I don't know that much about C. I'll amend it @vanhauser-thc

hydra-http.c
```diff
@@ -34,7 +45,7 @@ int32_t start_http(int32_t s, char *ip, int32_t port, unsigned char options, cha
 
   buffer_size = strlen(header) + 500;
   if (!(buffer = malloc(buffer_size))) {
-    free(header);
+    freeM(header);
```
vanhauser-thc, line 48: this - and all other - would need to be `&header` instead of `header` (because the function is void **ptr)

hydra-http.c
```diff
@@ -86,8 +97,8 @@ int32_t start_http(int32_t s, char *ip, int32_t port, unsigned char options, cha
     fooptr = buffer2;
     result = sasl_digest_md5(fooptr, login, pass, buffer, miscptr, type, webtarget, webport, header);
     if (result == NULL) {
-      free(buffer);
-      free(header);
+      frefreeM(buffer);
```
vanhauser-thc: typo, cant compile

hydra-http.c
```diff
@@ -447,88 +454,131 @@ int32_t service_http_init(char *ip, int32_t sp, unsigned char options, char *mis
...
+            return -1;
+          }
 
-      for (int i = 0; i < strlen(status_code); i++) {
-        if (!isdigit(*(status_code + i))) {
-          hydra_report(stderr, "Invalid status code, eg: 200");
-          return -1;
+          match_status_code[pos++] = value;
+
+          if (pos > MAX_STATUS_CODE_SIZE) {
```
vanhauser-thc, line 501: needs to be `>=`

hydra-http.c
```diff
@@ -447,88 +454,131 @@ int32_t service_http_init(char *ip, int32_t sp, unsigned char options, char *mis
...
+        }
+      }
+      if (strstr(p, "F=") != NULL || strstr(p, "S=") != NULL) {
+        int size = 0;
+        if (misc != NULL) {
+          size += strlen(misc) + 1;
         }
+        size += strlen(p);
+        match_text_start = (char *)malloc(size);
+        memset(match_text_start, '\0', strlen(match_text_start));
```
vanhauser-thc, line 525: memset unneeded here

hydra-http.c
boynicholas:
```suggestion
    void freeMemory(void **ptr) {
  if (ptr != NULL && *ptr != NULL) {
    free(*ptr);
    *ptr = NULL;
  }
}
```

Maybe I think that's correct here, with the processing inside the function body

boynicholas:
@vanhauser-thc I have tested it to achieve the requirements, thanks for providing it

hydra-http.c
vanhauser-thc: have you tested it? it should crash :) 
just try `char *foo = malloc(100); freeM(foo);`

boynicholas: If I use &header, it will report an error
```suggestion
    freeM(&
<img width="612" alt="20210707-231705@2x" src="https://user-images.githubusercontent.com/6275408/124785501-7ad8d680-df79-11eb-9df5-3eb043b97965.png">
header);
```

boynicholas: > have you tested it? it should crash :)
> just try `char *foo = malloc(100); freeM(foo);`

yes, I tried it and it didn't crash

boynicholas: Maybe i think you overlooked it
```c
void freeMemory(void **ptr) {
  if (ptr != NULL && *ptr != NULL) {
    free(*ptr);
    *ptr = NULL;
  }
}

#define freeM(p) freeMemory((void **)(&(p)))
```

hydra-http-form.c
```diff
@@ -509,8 +510,17 @@ int32_t parse_options(char *miscptr, ptr_header_node *ptr_head) {
       hydra_report(stderr, "[ERROR] Out of memory for HTTP headers (H).\n");
       return 0;
     default:
-      hydra_report(stderr, "[ERROR] no valid optional parameter type given: %c\n", miscptr[0]);
-      return 0;
+      while (*ptr != 0 && *ptr != ':')
```
vanhauser-thc, line 513: if an option is unknown an error should be printed and exited dunno why you changed the behaviour

hydra-http.c
vanhauser-thc: yeah I had overlooked that :)

hydra-http.c
```diff
@@ -10,9 +10,21 @@ char *http_buf = NULL;
 static char end_condition[END_CONDITION_MAX_LEN];
 int end_condition_type = -1;
 
+#define MAX_STATUS_CODE_SIZE 10
+int match_status_code[MAX_STATUS_CODE_SIZE];
+
```
vanhauser-thc, line 15: add `int match_status_code_cnt;`

hydra-http.c
```diff
@@ -248,6 +259,38 @@ int32_t start_http(int32_t s, char *ip, int32_t port, unsigned char options, cha
   ptr = ((char *)strchr(http_buf, ' '));
   if (ptr != NULL)
     ptr++;
+
+  // check status
+  if (match_status_code != NULL && match_status_code[0] != 0) {
```
vanhauser-thc, line 264: match_status_code can never be NULL. switch this to:
`if (match_status_code_cnt) {`

hydra-http.c
```diff
@@ -248,6 +259,38 @@ int32_t start_http(int32_t s, char *ip, int32_t port, unsigned char options, cha
...
+    for (int i = 0; match_status_code[i]; i++) {
```
vanhauser-thc, line 265: switch this to
```
int i; // for systems not support C99
for (i = 0; i < match_status_code_cnt; ++i) {
```

hydra-http.c
```diff
@@ -426,54 +468,135 @@ int32_t service_http_init(char *ip, int32_t sp, unsigned char options, char *mis
...
+        }
+
+        char tmp;
+        int value, n;
+        int pos = 0;
+
+        p += 2;
+
+        while (p) {
+          n = sscanf(p, "%3d%c", &value, &tmp);
```
vanhauser-thc, line 507: you should remove the 3 of %3d, because otherwise you cannot match below if the value is larger than 1000

hydra-http.c
```diff
@@ -426,54 +468,135 @@ int32_t service_http_init(char *ip, int32_t sp, unsigned char options, char *mis
...
+        if (strlen(p) < 3) {
+          hydra_report(stderr, "Invalid status code, eg: 200 or 200|400.");
+          free(cp);
+          misc = NULL;
+          return -1;
+        }
+
+        char tmp;
+        int value, n;
+        int pos = 0;
```
vanhauser-thc, line 502: remove pos and replace all pos with match_status_code_cnt

hydra-http-form.c
boynicholas: The parameter will actually be validated in both modules. When I set r=200, this parameter is not in hydra-http-form.c, so I have to ignore it. It will be validated in hydra-http.c

vanhauser-thc: yes but I can also do x=fooo which is a non-existing parameter and there will be no information about this. also if you use r= in http-form there will not even be a warning/error that this option is not supported there.

big-main (approved):

xxxxxspooferxxxxx (approved):

vanhauser-thc:
@boynicholas are you aware that you have set this PR to "draft" so it cannot be merged?
Hence I assume you are still working on it.
and are you involved in that other pull request that just came in? It has this one included: https://github.com/vanhauser-thc/thc-hydra/pull/679

PR #675, opened 2021-08-05
telnet detects password retries with same user

horner:
Some systems dont ask for the user, and instead give the user multiple attempts attempts at the password before switching back to asking for the username.

```
Trying 192.168.100.11...
Connected to 192.168.100.11.
Escape character is '^]'.

login: user
Password: *******
login incorrect.
Password: ********
login incorrect.
Password: *****
login incorrect.
login:
```

This PR changes the telnet protocol to detect the password prompt and get the next password to try.  Without this PR, hydra would get stuck in a timeout reading from the socket.


vanhauser-thc:
LGTM thank you!
what a horror that telnet is still a thing :)

PR #709, opened 2022-01-09
Rename README to README.md

nityanand86876:
Now `README.md` looks beautiful.
Why would you change `readme.md` to `readme` ?
This will not process `.md` file and print it as plain text.
So i changed `README` to `README.md` to look it beautiful.


nityanand86876:
if you want to see how readme looks like visit `[HERE](https://github.com/nityanand86876/thc-hydra/blob/master/README.md)`

PR #718, opened 2022-01-31
configure: openssl / memcached build fix

dbungert:
On Debian/Ubuntu, compilation against openssl 3.0 causes a failure to
find INT_MAX, despite the openssl headers including limits.h.  However,
the fact that the libmemcached-dev package provides both
/usr/include/libmemcached{,-1.0} directories, both of which contain
memcached.h, mean that MCACHED_IPATH ends up set to the libmemcached-1.0
one, which contains a limits.h, which shadows /usr/include/limits.h.
Don't do that.

Not sure how this plays out in other build environments, so feel free to
decline or suggest changes.  Thanks!


vanhauser-thc:
thank you!

PR #723, opened 2022-02-11
hydra-gtk: add support for termux path

Yisus7u7:
Screenshot_20220211-112627

This small change makes Xhydra work smoothly in termux x11


vanhauser-thc:
thank you

PR #744, opened 2022-04-11
Fix memory leak in radmin2

Teemperor:
`msg` is calloc'd a few lines above via `msg = buffer2message(buffer);`.
The check afterwards either exits the process on success or restarts the
loop without free'ing `msg`.


vanhauser-thc:
thank you!

Issue #806, opened 2022-12-04
rdp compatibility issue 00020008


ring04h:
libfreerdp > 2.1 version, default set tls-seclevel level to 1, the default security option is too high

hydra -d -l administrator -p 'aliyun123' 10.0.0.251 rdp

when i try to target Windows Server 2008 R2, get error 00020008

```
[DEBUG] rdp reported 00020008
[ERROR] freerdp: The connection failed to establish.
[DEBUG] pid 29636 called child_exit with code 1
```

```
add support to set tls security level (for openssl >= 1.1.0)
  - default level is set to 1
  - the new command line option /tls-seclevel:[LEVEL] allows to set a different level if required

Since Server 2008 R2 is no longer generally supported, I wouldn't
recommend decreasing the default seclevel (unless there are other
supported RDP server versions which require it) but it would be nice
if the error message gave users some suggestions for likely causes of
ERRCONNECT_TLS_CONNECT_FAILED and how to address them.

```

Suggest:
hydra set tls security level option to 0, like this: 

```
xfreerdp /tls-seclevel:0 /v:10.0.0.251
```

Reference:
https://github.com/FreeRDP/FreeRDP/blob/c5e425242a6b52cb4a70db09ba0afe9bd7b91087/libfreerdp/core/settings.c#L711

ring04h:
https://github.com/vanhauser-thc/thc-hydra/blob/master/hydra-rdp.c#L31

```c
instance->settings-> TlsSecLevel = 0;
```

vanhauser-thc:
fixed in #807 thanks

PR #807, opened 2022-12-04
freerdp tls-seclevel to 0

ring04h:
fix:
freerdp default set tls-seclevel level to 1, connection failed to establish

https://github.com/vanhauser-thc/thc-hydra/issues/806


vanhauser-thc:
yes that looks like the right move! thanks for the PR

PR #817, opened 2023-01-18
Various C99 compatibility fixes

fweimer-rh:
strrchr is declared in <strings.h>, not <string.h>.

_GNU_SOURCE needs to be defined before any glibc headers are included, otherwise it is not effective.

Also patch some old autoconf-internal issues in the hydra-gtk configure script.

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC


vanhauser-thc:
thank you!

PR #860, opened 2023-06-27
fix: array proxy_string_port may out of bound

neo-one0873:
when proxy_count=64, array proxy_string_port , proxy_string_ip, etc. may out of bound.


vanhauser-thc:
thanks!

PR #861, opened 2023-06-27
fix memory leaks for hydra-http-form.c

bearxy123:
fix memory leaks for cookie_header,you should free cookie_header before return


vanhauser-thc:
thanks!

Issue #863, opened 2023-06-28
http-post-form: H=Cookie\: password=^PASS^ creates a cookie with unexpected name and value


CoenTempelaars:
**Describe the bug**
http-post-form optional parameter `H=Cookie\: password=^PASS^` creates a cookie with:

- name: the password that `^PASS^` evaluates to
- value: empty

I would expect:

- name: the literal `password`
- value: the password that `^PASS^` evaluates to

My expectation is based on the help text of http-post-form, but in my humble opinion it is also common sense.

**To Reproduce**
Steps to reproduce the behavior: (I'm on version 9.4 but the bug is not mentioned in the 9.5 changelist)
1. Spin up a php server serving the document below
2. Attack the php server with hydra: `hydra -d -l admin -p "mypassword" -m "/:fname=^USER^:H=Cookie\: password=^PASS^:failed" http-post-form://127.0.0.1`
3. Observe that hydra defines cookie: `Cookie: mypassword`
4. Observe that the php server returns: `[mypassword] => .`

**Expected behavior**
In step 3 I would expect that hydra defines cookie: `Cookie: password=mypassword`

so that the php server returns in step 4: `[password] => mypassword.`

**Desktop (please complete the following information):**
 - OS: Kali Linux 2022.4
 - hydra version 9.4

**Additional context**
```
<html>
<body>

<form method="post" action="<?php echo $_SERVER['PHP_SELF'];?>">
  Name: <input type="text" name="fname">
  <input type="submit">
</form>

<?php
print_r($_COOKIE);
?>

</body>
</html>
```

CoenTempelaars:
I'll take that challenge to go for a fix and send a PR.

I am debugging hydra-http-form.c in gdb to figure out where to apply my fix.
I learned about gdb setting `follow-fork-mode child` yesterday.

If you have any hints about debugging thc-hydra and/or where to apply the fix then please let me know.

vanhauser-thc:
cool!

vanhauser-thc:
well it needs to be in the location where ^PASS^ is being replaced so there can only be very few locations for that in the file. I would put some fprintfs in there to see where this is going wrong

PR #867, opened 2023-07-08
fix replacement of user/pass placeholders in http header

CoenTempelaars:
Proposed solution for https://github.com/vanhauser-thc/thc-hydra/issues/863


vanhauser-thc:
Looks good, thanks!

PR #882, opened 2023-08-10
Allow HTTP-POST with F=403

danielvandenberg95:
I had a site which returns 200OK, but a json containing 403. Get results in "invalid api call". Allow using F= with post.


vanhauser-thc:
thanks

PR #898, opened 2023-09-28
Update hydra-http-form.c

andraxin:
Fix handling web forms that may return 401.

PR #912, opened 2023-12-03
Fix "make" error

Hatsumi-FR:
Variable 'tmp' was moved to a higher scope in the parse_options function of hydra-http-form.c. This change was necessary to prevent duplicate declarations in the different switch case blocks. This PR fix "make" error : 
```
hydra-http-form.c: In function ‘parse_options’:
hydra-http-form.c:451:13: error: redefinition of ‘tmp’
  451 |       char *tmp = strchr(miscptr, ':');
      |             ^~~
hydra-http-form.c:443:13: note: previous definition of ‘tmp’ with type ‘char *’
  443 |       char *tmp = strchr(miscptr, ':');
      |             ^~~
make: *** [Makefile:81 : hydra-http-form.o] Erreur 1
```

### DIFF
--- a/hydra-http-form.c
+++ b/hydra-http-form.c
@@ -390,7 +390,7 @@ char *stringify_headers(ptr_header_node *ptr_head) {
 }
 
 int32_t parse_options(char *miscptr, ptr_header_node *ptr_head) {
-  char *ptr, *ptr2;
+  char *ptr, *ptr2, *tmp;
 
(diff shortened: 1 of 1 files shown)


Hatsumi-FR:
it seems that this bug was implemented here: https://github.com/vanhauser-thc/thc-hydra/pull/898/files

vanhauser-thc:
LGTM, thanks

PR #914, opened 2023-12-05
fix smb password expired vs account expired confusion

tothi:
- If SMB password is MUST_CHANGE or EXPIRED, it is still valid, it is possible to change it remotely, so it should be returned as valid.
- If the _account_ is EXPIRED, it is no longer usable, even if the password is ok. It should be returned as invalid.

Testing is in this thread: https://twitter.com/an0n_r0/status/1731109539204710416

Originally Hydra returned valid if the account was expired but invalid if the password was expired. It is wrong, it should be exactly the opposite: valid if password was expired, but invalid if account was expired.

### DIFF
--- a/hydra-smb.c
+++ b/hydra-smb.c
@@ -1304,13 +1304,12 @@ int32_t start_smb(int32_t s, char *ip, int32_t port, unsigned char options, char
       hydra_report(stderr, "[INFO] LM dialect may be disabled, try LMV2 instead\n");
     hydra_completed_pair_skip();
   } else if (SMBerr == 0x000024) { /* change password on next login [success] */
-    hydra_report(stdout, "[%d][smb] Host: %s Account: %s Error: ACCOUNT_CHANGE_PASSWORD\n", port, ipaddr_str, login);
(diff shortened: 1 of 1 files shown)


hydra-smb.c
```diff
@@ -1311,7 +1311,7 @@ int32_t start_smb(int32_t s, char *ip, int32_t port, unsigned char options, char
   } else if (SMBerr == 0x000071) { /* password expired */
     if (verbose)
       fprintf(stderr, "[%d][smb] Host: %s Account: %s Error: PASSWORD EXPIRED\n", port, ipaddr_str, login);
```
vanhauser-thc, line 1313: I think "Error" in that line is then not helpful? Maybe change it to "Information"?

tothi: Absolutely, "Information" is much better.

vanhauser-thc:
@tothi can you please fix the "Error" string to "Information" so I can merge this? thanks you

PoesJakals:
Carynw_ Instagram password

tothi:
added some logging fix. sorry for the super long delay, totally forgot it :)

PR #915, opened 2023-12-05
add legacy SSH ciphers support

tothi:
Legacy SSH ciphers (for Key Exchange and Host Keys) was missing from the default cipher list, it is needed to specify it in custom libssh options. Here are some common useful extra legacy ciphers added.

NOTE: Currently this hardcoded list is working, but for the future it might be better to _add_ the legacy ciphers to the default ones, not defining the full list (if some new ciphers will be added to the defaults automatically in libssh).


vanhauser-thc:
lgtm, thanks

PR #947, opened 2024-05-07
Bump Debian version in Dockerfile to Bookworm 

Proto1337:
Since Debian Buster's LTS support ends next month, I suggest updating the base image in the Dockerfile to Debian Bookworm.


vanhauser-thc:
thanks!

PR #970, opened 2024-08-13
Fix compile errors when compiling against freerdp3

ccc-aaa:
FreeRDP/FreeRDP#7738
use settings field under `instance->context` instead of `instance`.

FreeRDP/FreeRDP@5f81005ecbd87d043f2ce39ea11c98065b1a1923
removes reference to MaxTimeInCheckLoop for version 3 since FreeRDP has dropped this field after migrating away from blocking poll loop.


vanhauser-thc:
weird I did not get a notification, so this fell off my radar ... sorry!

PR #997, opened 2025-01-26
Update README

sadab82984529:
Number one of the biggest security holes are passwords, as every password
security study shows.
This tool is a proof of concept code, to give researchers and security
consultants the possibility to show how easy it would be to gain unauthorized
access from remote to a system.


ww1007jsj (approved):

ww1007jsj:
Hhsj

PR #1007, opened 2025-03-19
Use library for mssql

### LINKED ISSUE #814: MSSQL module has an undocumented username and password length limit
The MSSQL module has an undocumented and silent username and password limit of 30 characters.

See here: https://github.com/vanhauser-thc/thc-hydra/blob/master/hydra-mssql.c#L68-L71

It just chops the password off without any warning to the user. I've not yet tried recompiling the code with a higher limit (maybe there are other protocol reasons for this limit?) but that might be worth trying for anyone who needs to explore brute forcing characters longer than 30 characters on MSSQL.

motypi:
resolves #814 
* new start_mssql function using TDSv7
* keeps support for TDSv4


hydra-mssql.c
```diff
@@ -1,10 +1,48 @@
...
+
+  if (strlen(login = hydra_get_next_login()) == 0)
+    login = empty;
+  if (strlen(pass = hydra_get_next_password()) == 0)
+    pass = empty;
+
+  DBPROCESS *dbproc;
+  LOGINREC *attempt;
+
+  dbinit();
```
vanhauser-thc, line 21: I do not the library but my guess is that either dbinit() is only required once instead of every login attempt, or there is a dbclose() or similar function that needs to be called after every login attempt. please check.

motypi: There is indeed a dbclose(), and I added it. Thanks for the feedback

vanhauser-thc:
Good that you added the dbclose etc. - but are they needed?
Did you test this?
because it could be that you just need once the dbinit and then can try to auth endless times and be much faster. or maybe there is a limit. or you can only try once ... so please test and report :-)

motypi:
Hey Marc, I did speed tests and I realise that 10x slower than the previous version. Sorry I hadn't tested speed previously. I implemented the change you proposed, but the issue is that the client is sending pre-login requests and doing a handshake with every login attempt, which happens during dbopen(). It seems like the solution should be for me to find a way to authenticate without the handshakes, which doesn't seem to be an option provided by the API. Maybe we could close for now and I come back another time with a workaround?

motypi:
Hi Marc, I made it so that only passwords longer than 30 characters are tested using TDS7. That helps with the speed issues!

PR #1019, opened 2025-06-03
Fix several memory leaks

oss-belobog:
Hello!

Thanks for the fantastic project.

We found a few memory leaks during our code review and this patch fixes them by correctly releasing the buffer.


hvik01 (approved):

vanhauser-thc:
thanks!

PR #1047, opened 2025-11-21
More hydra gtk

elboulangero:


Makefile.am
```diff
@@ -59,10 +59,10 @@ hydra:	hydra.c $(OBJ)
 	@echo
 
 xhydra:	
-	-cd hydra-gtk && sh ./make_xhydra.sh
+	cd hydra-gtk && sh ./make_xhydra.sh
```
vanhauser-thc: if xhydra cannot be built then this should not result in a build failure. not everyone has gtk installed.

vanhauser-thc:
I noticed that the autconfigure stuff now modifies the `hydra-gtk/INSTALL` file

elboulangero:
`hydra-gtk/INSTALL` yes it's a auto-generated file, I thought that it would be nice to keep it committed to the repo, in case someone wants to read it.

But then depending on what version of `autoreconf` you run, it will probably be slightly different...

So maybe best to remove it from the repo like all the other files? Agree with that?

PR #1053, opened 2025-12-17
Enhance password-only handling in Telnet service

s-b-repo:
Key ImprovementsAutomatic detection of password-only mode: After Telnet negotiation, checks banners for "password:" (or variants) without "login:" or "username:". Skips sending username if password-only detected — directly sends the password. More robust prompt detection: Added "pin:" and better draining of initial banners. Fallback safe: If server re-prompts for login after bad password, it correctly skips the pair. Works with empty login: Still supports manual -l "" for compatibility.

PR #1058, opened 2026-01-23
Fix: telnet compilation error in hydra-telnet.c from PR #1053

Jettlez:
Fixes the compile errors from the #1053 merge of **hydra-telnet.c**.

The new **hydra-telnet.c** file removed the **arpa/telnet.h** header; however functionality of the file was never added back, and protocols like (IAC, WILL, DONT, etc) and telcmds functions were missing. This just adds the protocols in a new file **hydra-telnet.h** and allows the telnet service to compile properly.

Changes:
- Added back Telnet constants defined in arpa/telnet.h to hydra-telnet.h.
- moved array definition to the main .c file and used hydra-telnet.h file to make it cleaner.
- fix allowed compilation on Linux "6.18.4-1-MANJARO".

### DIFF
--- a/hydra-telnet.c
+++ b/hydra-telnet.c
@@ -1,5 +1,12 @@
 #include "hydra-mod.h"
+#include "hydra-telnet.h"
 #include <ctype.h>
+
+char *telcmds[] = {
+  "EOF", "SUSP", "ABORT", "EOR",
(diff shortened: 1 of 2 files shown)


vanhauser-thc:
thanks!

PR #1062, opened 2026-02-14
Fix SSH port from CLI being overridden by ~/.ssh/config

sho29:
Fixing issue recommended by issue#1039

1. Changes made to two files that call ssh_options_parse_config() before ssh_options_set so that command-line port takes priority over config file defaults.
2. libssh's ssh_connect() auto parses config files and overrides previously set options. By explicitly parsing first, then setting the port, the CLI value wins.


vanhauser-thc:
thanks

PR #1078, opened 2026-05-25
Fix premature success detection in telnet module

ebrasha:
Noticed the Telnet module was producing a ton of false positives. The core problem was that the original code treated shell prompt characters (`/`, `>`, `%`, `$`, `#`) anywhere in a server response as proof of a successful login, even before credentials were sent.

The original had two "early success detection" blocks in `start_telnet()`: one in the banner-reading loop, and another right after sending the username (before the password). If a banner like "Welcome to Cisco IOS, press # for help" or "Server v1.0, 100% reliable" came back, hydra called `hydra_report_found_host` immediately. Same thing happened post-username: a reply like "Password (use # to cancel):" got reported as a win.

The original `is_failure()` also matched bare tokens like "wrong", "failed", and "too many" with case-sensitive `strstr`. Those show up all over MOTDs and help text, and the case-sensitive matching meant mixed-case failure messages slipped through entirely.

**What I changed:**

* Stripped out the early success-detection blocks completely. Success is never reported before the password gets sent.
* Wrote `has_shell_prompt()`, which only checks the last non-whitespace byte of the response for `$`, `#`, `>`, `%`. That still catches real shell prompts (bash, csh, Cisco IOS) but ignores paths, percentages, and help text mid-buffer.
* Tightened `is_failure()` to match multi-word phrases only ("bad password", "access denied", "login incorrect", "too many attempts", etc.) and made it always lower-case the input so case doesn't matter.
* Split prompt detection into `is_password_prompt()`, `is_login_prompt()`, and `is_login_prompt_banner()` helpers, with the login prompt detection deliberately stricter post-auth so MOTD text doesn't get mistaken for a re-prompt.

### DIFF
--- a/hydra-telnet.c
+++ b/hydra-telnet.c
@@ -268,9 +268,10 @@ static telnet_resp_t classify_response(const char *buffer) {
  * A literal ':' inside a value can be escaped as "\:" exactly as in the
  * http-form module, so success strings such as "Last login\:" are usable.
  *
- * The caller is expected to have already lower-cased miscptr; this routine
- * only splits and unescapes. The function is idempotent and safe to call
- * multiple times; previous values are released first.
+ * The prefix detection is case-insensitive so that both "S="/"F=" and the
+ * already-lower-cased "s="/"f=" produced by the call-site work the same
+ * way. The function is idempotent and safe to call multiple times;
+ * previous values are released first.
  */
 static void telnet_parse_miscptr(char *miscptr) {
   char *work, *seg_start, *p, *unescaped;
@@ -288,7 +289,7 @@ static void telnet_parse_miscptr(char *miscptr) {
   if (miscptr == NULL || *miscptr == 0)
     return;
 
-  has_prefix = (strncmp(miscptr, "s=", 2) == 0 || strncmp(miscptr, "f=", 2) == 0);
+  has_prefix = (strncasecmp(miscptr, "s=", 2) == 0 || strncasecmp(miscptr, "f=", 2) == 0);
 
   if (!has_prefix) {
     /* Legacy contract: whole value is a success substring. */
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
That looks useful.

A caveat though:
> Success is never reported before the password gets sent.

Some systems where a user has no password set, no password prompt is set and the user is directly logged in.
This must still be detected correctly.

Another thing I wanted to add some point but never did: allow the user to specify a "fail string" via the optional module specific parameter

ebrasha:
> That looks useful.
> 
> A caveat though:
> 
> > Success is never reported before the password gets sent.
> 
> Some systems where a user has no password set, no password prompt is set and the user is directly logged in. This must still be detected correctly.
> 
> Another thing I wanted to add some point but never did: allow the user to specify a "fail string" via the optional module specific parameter


👍 **Thanks, both points are fair.**

You were right about the passwordless case — I was too aggressive with the "never report success before the password is sent" rule and ended up breaking detection of accounts that don't have a password set. Fixed it: after the username is sent, the loop now classifies the response the same way the post-password loop does (failure phrase → login re-prompt → password prompt → trailing shell prompt). If the server skips the password step and drops us straight into a shell, we report the current pair as found, same as the legacy module did. The pre-credential banner phase still never declares success, since at that point we genuinely have nothing to go on yet.

For the fail-string idea, I went ahead and added it while I was in there. Reused the same `S=` / `F=` / `\:` convention that http-form already uses so there's nothing new to learn:

```
-m "string"                          legacy: whole value used as success substring
-m "S=Last login"                    explicit success substring
-m "F=Authentication failed"         extra fail substring, on top of the built-in list
-m "S=Last login:F=Auth failed"      both, order doesn't matter
\:                                   literal ':' inside a value
```

Both phases now go through a single `classify_response()` helper, so the operator-supplied fail string is honored in the post-username phase too, not just post-password. That helps with devices that use non-standard rejection messages — an operator who knows exactly what their target prints on a bad login can plug it in and stop relying on the built-in phrase list entirely.

Updated `usage_telnet()` to document the new syntax and the passwordless behavior. No signature changes; the `SERVICE3` entry stays the same.

vanhauser-thc:
looks good.
> expected S= or F=
in the code you check for "s="/"f=" (lowercase)

ebrasha:
> looks good.
> 
> > expected S= or F=
> > in the code you check for "s="/"f=" (lowercase)



Good catch, thanks. Swapped the three prefix checks from `strncmp` to `strncasecmp` so the parser now matches what the warning message and the usage text already say — `S=` / `F=` is accepted directly, no longer leaning on the caller having lower-cased miscptr beforehand. The lower-casing in `service_telnet()` is still there because the matched substrings need to be lower-case to compare against the lower-cased response buffer, but the prefix detection itself is now case-insensitive on its own, so it stays correct even if that pre-processing ever moves or changes.

PR #1079, opened 2026-05-26
Bugfix/http status false positives

ebrasha:
While going through false-positive on the HTTP module I noticed something that probably accounts for a chunk of them: the default success detection treats 2xx, 3xx, 403, and 404 all as "successful login" if you haven't supplied an explicit F= or S=. That's a pretty wide net, and three of those four ranges don't really say anything useful about the credentials.

A few concrete cases where it bites you:

- 403 Forbidden — some apps return this on bad credentials, others return it after a successful auth when you're hitting a resource you're not allowed on. Same status, opposite meaning.
- 404 Not Found — just means the URL doesn't exist. The credentials could be valid or not, the response doesn't care.
- 3xx — could be a redirect to /login?err=1 (you failed) or to /dashboard (you succeeded). The status alone tells you nothing.

401 is the one status that's unambiguous in this context (Basic/Digest/NTLM all use it for "wrong credentials, here's a challenge"), so the only response class that should imply a real success without operator help is the 2xx family: 200, 201, 204, 206, and so on.

So the default is now narrower. Only a 2xx response counts as success unless the operator opts in. An explicit S= or F= still wins over everything else, same as before.

The tricky case is redirects. There are perfectly normal sites where a 302 is the only success signal you get (the dashboard lives at /home and the login form just redirects you there on success). Making those users dig into the response body to write an S= would be annoying, so there's a new optional parameter R= that lets you state the policy:

```
R=failure              default. any 3xx counts as a failed attempt
R=success              flip it back to the old behaviour; any 3xx is success
R=location=/dashboard  3xx only counts as success if Location contains /dashboard
R=location\:/dashboard same thing with the colon escaped if you prefer that form
```

The location match only looks at the actual `Location:` header line (it's anchored to the start of a header, not a stray string in the body), and the comparison is case-insensitive.

I didn't add an equivalent toggle for 403 or 404 on purpose. Neither status carries enough meaning on its own to be worth a built-in switch, and if you really need them to count as success on some weird target the right move is to write an S= or F= for the body, not to trust the status code.

A few examples of what the new flow looks like:

Cisco-style appliance that drops you onto a config page after auth (a 302 to /home or similar):

```
hydra -l admin -P pwd.txt 10.0.0.1 http-get '/:A=BASIC:R=success'
```

Login page that 302s to /dashboard.aspx on success and back to /login on failure:

```
hydra -l admin -P pwd.txt example.com http-get '/login:R=location=/dashboard.aspx'
```

Target with a known post-login banner (still the most reliable mode and the one I'd reach for first when I know the body):

```
hydra -l admin -P pwd.txt example.com http-get '/admin:S=Welcome back, admin'
```

Auth landing page that returns 200 for both success and failure and only the body differs:

```
hydra -l admin -P pwd.txt example.com http-get '/admin:F=Invalid credentials'
```

Backwards compatibility: if you were relying on a 3xx-as-success default for some target you've been brute-forcing, you'll need to add `R=success` (or something more specific) to keep things working. Anything that was already driven by S= / F= keeps working without changes.

### DIFF
--- a/hydra-http-form.c
+++ b/hydra-http-form.c
@@ -499,6 +499,36 @@ int32_t parse_options(char *miscptr, ptr_header_node *ptr_head) {
       getcookie = 0;
       miscptr = ptr;
       break;
+    case 'r': // fall through
+    case 'R':
+      ptr = miscptr + 2;
+      while (*ptr != 0 && (*ptr != ':' || *(ptr - 1) == '\\'))
+        ptr++;
+      if (*ptr != 0)
+        *ptr++ = 0;
+      tmp = hydra_strrep(miscptr + 2, "\\:", ":");
+      if (tmp == NULL)
+        tmp = miscptr + 2;
+      if (strcasecmp(tmp, "success") == 0) {
+        redirect_condition_type = REDIRECT_CONDITION_SUCCESS;
+        redirect_condition[0] = 0;
+      } else if (strcasecmp(tmp, "failure") == 0) {
+        redirect_condition_type = REDIRECT_CONDITION_FAILURE;
+        redirect_condition[0] = 0;
+      } else if (strncasecmp(tmp, "location=", 9) == 0 || strncasecmp(tmp, "location:", 9) == 0) {
+        char *location_match = tmp + 9;
+        if (strlen(location_match) >= REDIRECT_CONDITION_MAX_LEN) {
+          hydra_report(stderr, "[ERROR] R=location value cannot be bigger than %u.\n", REDIRECT_CONDITION_MAX_LEN - 1);
+          return 0;
+        }
+        redirect_condition_type = REDIRECT_CONDITION_LOCATION;
+        strcpy(redirect_condition, location_match);
+      } else {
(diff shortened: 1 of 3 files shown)


vanhauser-thc:
Switching the default behaviour is not a good user experience IMHO

I think `R=success` should be the default, otherwise this might fail existing scripts, tests, setups, ...

ebrasha:
> Switching the default behaviour is not a good user experience IMHO
> 
> I think `R=success` should be the default, otherwise this might fail existing scripts, tests, setups, ...

Fair point. Default is now R=success, so historical 3xx behaviour is preserved and existing scripts keep working. R=failure / R=location=<text> are still there as opt-ins for stricter handling.

I left 403/404 as failure by default — those don't really prove a successful login on their own (403 in particular is ambiguous: some apps use it for bad creds, others for post-auth access denial).

vanhauser-thc:
Thanks!

ebrasha:
> Thanks!

You're welcome! It's an honor ❤️

PR #1083, opened 2026-06-24
rdp: avoid false positives on targets without NLA (issue #923)

### LINKED ISSUE #923: False positives in xrdp weak credential scanning
**Describe the bug**
When Hydra scans an xrdp service, it always reports any username/password pair used to be valid, while printing out an error `[ERROR] freerdp: The connection failed to establish.` at the same time (even with the correct credential). 

I've set up a Debian vm with xrdp. When I used Microsoft Remote Desktop to connect to it, the client behaviour was a bit unexpected (though I think it could be an x?rdp protocol quirk):

1. If the credentials are correct, I can directly log into my debian instance
2. If the credentials are incorrect, the initial connection is still established, then I get redirected to the xorg login portal:
Screenshot 2024-01-10 at 10 33 07 AM
Screenshot 2024-01-10 at 10 33 30 AM

I suspect this xrdp behaviour caused Hydra to always assume any credential pair is valid, because the initial connection is always established. 

**To Reproduce**

1. Enable xrdp on a linux vm: https://linuxize.com/post/how-to-install-xrdp-on-debian-10/
2. Scan that vm with hydra: `hydra -l root -p 'root' <linux_vm_ip> rdp`, using any username/password 

**Expected behavior**
Ideally Hydra reports valid credential only if it can actually log into the debian instance 
If xrdp is not officially supported, it would be great to have a way to detect and skip xrdp services so that Hydra doesn't generate false positive findings.  

**Desktop (please complete the following information):**
 - OS: Hydra is running in dockerized environment, using `openjdk:11-jdk-bullseye` as the base image, and installed via `apt-get install -y hydra`, which installed `libfreerdp2-2/now 2.3.0+dfsg1-2+deb11u1 amd64 [installed,local]` as part of the dependencies. 
 - hydra version v9.1

noderaven:
Fixes #923.

### The bug

The `rdp` module reports every username/password pair as valid when the target does not negotiate NLA/CredSSP (default xrdp, or a Windows host with NLA disabled). It treats any successful `freerdp_connect()` as a valid login, but on a non-NLA server authentication is deferred to an in-session login prompt, so an `AuthenticationOnly` connect succeeds for any credentials. This matches the analysis in the issue: `xfreerdp ... +auth-only` returns the same result for correct and incorrect passwords.

### Why not actually validate the login?

The ideal fix would confirm whether a login really succeeded on these servers. That is not practical:

- Without NLA there is no protocol-level credential check at all, so `+auth-only` cannot tell right from wrong passwords (as shown in the issue).
- The only remaining signal is visual: complete the full graphical session and classify the rendered screen, a desktop (success) versus the xrdp login/error window (failure). That means rendering and heuristically interpreting the framebuffer, which is fragile across xrdp versions, themes, and locales, and would add significant brittle code to the module.

### What this PR does

After a successful connect it checks the negotiated security protocol (`SelectedProtocol`):

- NLA/CredSSP (`PROTOCOL_HYBRID` / `PROTOCOL_HYBRID_EX`) used: the credentials were validated during the handshake, so it is a genuine hit. Behavior is unchanged for these targets.
- NLA not used: the credentials were never validated, so it prints a clear message and disables the target (`hydra_child_exit(3)`) instead of emitting false positives.

`PROTOCOL_HYBRID*` live in a private FreeRDP header, so the two values used are defined locally. The module usage text now documents the NLA requirement.

### Why this is the right tradeoff

- It eliminates the false positives reliably, which is the actual problem reported, and it is the "detect and skip xrdp" behavior the reporter explicitly accepted as a good outcome.
- It is correct in general, not an xrdp-specific hack: it also stops the same false positives against NLA-disabled Windows hosts.
- It is minimal and leaves the working NLA path untouched, so there is no regression risk for the targets the module is actually tested against.

### Testing

- Compile-verified clean against FreeRDP 3.5.1, including `-Wall -Wextra` (no warnings on the changed code); the full `hydra` binary links and runs, and `hydra -U rdp` shows the new note.
- Runtime repro for reviewers:
  - Non-NLA (the bug): run an xrdp instance, for example via https://github.com/satishweb/docker-xrdp, then `hydra -l anyuser -p anypass <host> rdp`. Before: a "valid" result for every pair. After: `[ERROR] <host> does not enforce NLA/CredSSP for RDP ... Disabling this target.` and no false positive.
  - NLA Windows (no regression): correct credentials still report as found; wrong credentials report none.

### DIFF
--- a/hydra-rdp.c
+++ b/hydra-rdp.c
@@ -17,6 +17,21 @@ void dummy_rdp() { printf("\n"); }
 
 #include <freerdp/freerdp.h>
 #include <freerdp/version.h>
+
+/* RDP security protocols selected during the X.224 negotiation
+ * (see [MS-RDPBCGR] 2.2.1.2.1). They are defined in a private FreeRDP header
+ * (libfreerdp/core/nego.h), so the values we need are redefined here.
+ * Only NLA/CredSSP (PROTOCOL_HYBRID*) validates the supplied credentials during
+ * the connection handshake. Without it an AuthenticationOnly connect succeeds
+ * for any credentials, which is the root cause of the xrdp false positives in
+ * issue #923. */
+#ifndef PROTOCOL_HYBRID
+#define PROTOCOL_HYBRID 0x00000002
+#endif
+#ifndef PROTOCOL_HYBRID_EX
+#define PROTOCOL_HYBRID_EX 0x00000008
+#endif
+
 freerdp *instance = 0;
 BOOL rdp_connect(char *server, int32_t port, char *domain, char *login, char *password) {
   int32_t err = 0;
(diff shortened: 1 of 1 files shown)


vanhauser-thc:
that looks useful.

can you try to move this to the service_rdp_init function? because this is where such a check belongs (see hydra-ssh.c for example) - return 1 if the target service not working correctly, 0 if everything is fine.

noderaven:
Makes sense. Moved it into `service_rdp_init`, same approach as `service_ssh_init`: one probe before the children fork (dummy creds, since the security layer is negotiated before auth), returning 1 when there's no NLA/CredSSP, 3 if it can't connect, 0 otherwise. Dropped the old per-attempt code in `start_rdp`. Compile-tested against FreeRDP 3.5.1, and the docker-xrdp repro in the description still applies.

vanhauser-thc:
LGTM thanks

Issue #673, opened 2021-07-30
[STATUS] attack finished for smtp.gmail.com (waiting for children to complete tests) [VERBOSE] Disabled child 0 because of too many errors [VERBOSE] Disabled child 7 because of too many errors [VERBOSE] Disabled child 14 because of too many errors [VERBOSE] Disabled child 4 because of too many errors [VERBOSE] Disabled child 5 because of too many errors [VERBOSE] Disabled child 10 because of too many errors [VERBOSE] Disabled child 12 because of too many errors 1 of 1 target completed, 0 valid password found [WARNING] Writing restore file because 6 final worker threads did not complete until end. [ERROR] 6 targets did not resolve or could not be connected [ERROR] 0 target did not complete Hydra (https://github.com/vanhauser-thc/thc-hydra) finished at 2021-07-30 07:46:50


nitinkumaralwariya:
**IMPORTANT**
hydra not crack gmail passowrd

PR #1104, opened 2026-07-28
Fix bogus "did not resolve or could not be connected" and early exit with live children

### LINKED ISSUE #1102: `[ERROR] N targets did not resolve or could not be connected` printed after a fully successful attack (k is reused for the active-head count)
**Describe the bug**

After an attack that resolves, connects and finds a valid pair, hydra can still print
`[ERROR] N targets did not resolve or could not be connected` and exit 255.

The count in that message is not a target count. At `hydra.c:4430` the variable `k` — computed
at `hydra.c:4400-4423` from target state — is overwritten with the number of worker heads still
in `HEAD_ACTIVE`, and the message at `hydra.c:4475` is printed from the overwritten value. The
`[WARNING]` line at `hydra.c:4446` prints the same variable and labels it correctly, so a single
run reports one quantity under two different meanings, 29 lines apart.

The output is self-contradictory: `1 of 1 target successfully completed` and `1 target did not
resolve or could not be connected` appear together.

**To Reproduce**

Tested against current github master (`4b552dc`, v9.8dev), not a distro package.

The trigger is a race, so it needs concurrency: it shows up when several hydra processes run at
once against a fast-completing target, so that child exits bunch together and overlap the main
loop's teardown. A single sequential run rarely hits it.

1. Start a target that answers instantly:

   ```
   docker network create hydrarepro
   docker run -d --name pg --network hydrarepro --network-alias postgres.local \
     -e POSTGRES_PASSWORD=password postgres:15.7
   ```

2. Build hydra from master:

   ```
   git clone --depth 1 https://github.com/vanhauser-thc/thc-hydra /src
   cd /src && ./configure && make
   ```

3. Full command line used for every process:

   ```
   hydra -I -C /tmp/pg.txt -t 4 postgres://postgres.local
   ```

   where `/tmp/pg.txt` is a small colon file whose valid pair is near the top:

   ```
   dcmadmin:passw0rd
   postgres:amber
   postgres:postgres
   postgres:password
   ```

4. Run it 16-way concurrent, 20 rounds, and count:

   ```sh
   #!/bin/sh
   bogus=0; total=0; ok=0
   for r in $(seq 1 20); do
     i=1
     while [ "$i" -le 16 ]; do
       hydra -I -C /tmp/pg.txt -t 4 postgres://postgres.local > "/tmp/l.$i" 2>&1 &
       i=$((i+1))
     done
     wait
     i=1
     while [ "$i" -le 16 ]; do
       total=$((total+1))
       grep -q '1 valid password found' "/tmp/l.$i" && ok=$((ok+1))
       grep -q 'did not resolve or could not be connected' "/tmp/l.$i" && bogus=$((bogus+1))
       i=$((i+1))
     done
   done
   echo "runs=$total valid_password_found=$ok bogus_footer=$bogus"
   ```

Results on a 15-core machine, all runs against the same live postgres:

| build | concurrency | runs | valid password found | bogus footer |
|---|---|---|---|---|
| master `4b552dc` | 16 | 320 | 320 | **25** (7.8%) |
| v9.7 | 9 | 180 | 180 | **1** (0.6%) |

Every affected run found the password. Exit code on affected runs is **255** (checked over 10
occurrences).

**Expected behavior**

A run that resolved its target, connected, and found a valid pair should not report a
resolution/connection error, and should exit 0. If the intent is to warn that a worker head was
still active at exit, that is already covered by the `[WARNING] ... final worker threads did not
complete until end.` line.

**Screenshots**

Terminal output of an affected run (master `4b552dc`):

```
[5432][postgres] host: postgres.local   misc: (null)   login: postgres   password: password
1 of 1 target successfully completed, 1 valid password found
[WARNING] Writing restore file because 1 final worker threads did not complete until end.
[ERROR] 1 target did not resolve or could not be connected
[ERROR] 0 target did not complete
Hydra (https://github.com/vanhauser-thc/thc-hydra) finished at 2026-07-27 17:36:16
```

With `-b jsonv1` on v9.7, the same run writes the found credential and still marks the run
failed:

```json
"results": [ {"port": 5432, "service": "postgres", "host": "postgres.local",
              "login": "postgres", "password": "password"} ],
"success": false,
"errormessages": [ "[ERROR] 1 target did not resolve or could not be connected",
                   "[ERROR] 0 target did not complete" ],
"quantityfound": 1
```

**Desktop:**
 - OS: Ubuntu 24.04 (container), host Darwin 25.5.0, 15 cores
 - hydra version: current github master `4b552dc` (v9.8dev); also reproduced on v9.7

**Additional context**

Two separate defects combine here. Line numbers are v9.7, with master equivalents in parentheses.

*1. `k` is reused.* `hydra.c:4400-4423` counts unresolved/unconnectable targets. `hydra.c:4430`
(master 4512) throws that away:

```c
  error += j;
  k = 0;
  for (i = 0; i < hydra_options.max_use; i++)
    if (hydra_heads[i]->active == HEAD_ACTIVE)
      k++;
```

`hydra.c:4446` (master 4581) prints it as worker threads — correct. `hydra.c:4475` (master 4610)
prints it as targets — incorrect.

*2. Why a head is still active at exit.* `hydra_kill_head()` decrements the global counter
without checking whether the head was actually active, while the head's own state transition is
guarded:

```c
  if (killit) {
    if (hydra_heads[head_no]->pid > 0)
      kill(hydra_heads[head_no]->pid, SIGTERM);
    hydra_brains.active--;                        // 1425 (master 1521) - unguarded
  }
  if (hydra_heads[head_no]->active == HEAD_ACTIVE) {
    hydra_heads[head_no]->active = HEAD_UNUSED;   // 1428 - guarded
```

That function is called both from the main loop and from the async `SIGCHLD` handler
`killed_childs()` (`hydra.c:1030`, master 1126; installed for the mother at `hydra.c:3956`),
which reaps a child and calls `hydra_kill_head(i, 1, 0)` from signal context. `hydra_brains` is
a plain global — there is no `volatile` in `hydra.c`. When both paths reap the same child, one
increment gets two decrements. Once `hydra_brains.active` under-runs,
`hydra_check_for_exit_condition()` returns 1 while heads are still running:

```c
  if (hydra_brains.targets <= hydra_brains.finished && hydra_brains.active < 1) {   // 2064
    return 1;
```

The loop leaves, the recomputed `k` is non-zero, `error` is set, and `hydra.c:4511` returns -1.

This also explains #673 (2021), which was closed without a fix and shows the same signature from
a single-target run: `1 of 1 target completed` followed by `[WARNING] ... 6 final worker threads
did not complete until end.` and `[ERROR] 6 targets did not resolve or could not be connected`.

*Suggested fixes*

1. Use a separate variable for the active-head count at `hydra.c:4430` so `hydra.c:4475` keeps
   reporting the target-derived value, or reword 4475 to match what it now counts. Happy to send
   a PR for this part.

2. Pair the decrement with the state it tracks:

   ```c
   if (killit) {
     if (hydra_heads[head_no]->pid > 0)
       kill(hydra_heads[head_no]->pid, SIGTERM);
     if (hydra_heads[head_no]->active == HEAD_ACTIVE)
       hydra_brains.active--;
   }
   ```

   This is the part that causes the early exit. Since the state is shared with an async signal
   handler you may prefer a different approach — deferring the handler's work to a flag consumed
   by the main loop, and/or making the shared counters `volatile sig_atomic_t` — so I have not
   guessed at the design.

*Unrelated observation while testing:* on master `4b552dc`, `-b jsonv1 -o <file>` fails with
`[ERROR] Error creating outputfile: Invalid argument`, where v9.7 writes the file normally. Not
part of this report; happy to open it separately if useful.

gnuletik:
Fixes #1102 (and #673, which shows the same signature). Thanks for taking a look — sending the PR as requested.

Two defects combine. The first causes the early exit, the second misreports it.

### 1. `hydra_brains.active` can be decremented twice for one head

`hydra_kill_head()` decremented the counter without checking that the head was counted in it, while the head's own state transition immediately below *was* guarded:

```c
  if (killit) {
    if (hydra_heads[head_no]->pid > 0)
      kill(hydra_heads[head_no]->pid, SIGTERM);
    hydra_brains.active--;                        // unguarded
  }
  if (hydra_heads[head_no]->active == HEAD_ACTIVE) {
    hydra_heads[head_no]->active = HEAD_UNUSED;   // guarded
```

The function is called both from the main loop and from the async `SIGCHLD` handler `killed_childs()`, which reaps a child and calls `hydra_kill_head(i, 1, 0)`. When both paths reach the same head, one increment gets two decrements. Once the counter falls below the number of running heads, `hydra_check_for_exit_condition()` takes the `targets <= finished && active < 1` branch and the attack ends while children are still alive — which is how a head is left in `HEAD_ACTIVE` at the end.

The patch decrements only for a `HEAD_ACTIVE` head.

I deliberately did not touch the signal-handler design. Deferring the handler's work to a flag consumed by the main loop, and/or making the shared counters `volatile sig_atomic_t`, would be a larger change and yours to direct — the guard is enough to close the observed race. Happy to follow up if you want that too.

### 2. `k` was reused for two different quantities

`k` is computed from target state, then overwritten with the count of heads still in `HEAD_ACTIVE`. The `[WARNING]` restore-file line prints it as worker threads, which is correct. The `[ERROR]` line 29 lines further down prints the same variable as targets, which is not:

- a run with a leftover head reported `[ERROR] N targets did not resolve or could not be connected` even though every target resolved, connected and yielded a valid pair;
- a run whose target genuinely did not resolve reported **nothing** in the footer and exited 0, because `k` had been zeroed.

The patch gives the head count its own variable and leaves `k` as the target-derived value.

### Measurements

16 concurrent processes against a live `postgres:15.7`, `-I -C <4-line colon file> -t 4`, on a 15-core machine. 640 runs per build, same container, both binaries built from the same tree:

| build | runs | credential found | bogus footer | non-zero exit |
|---|---|---|---|---|
| `4b552dc` (master) | 640 | 634 | **15** | **15** |
| this branch | 640 | **640** | 0 | 0 |

Worth calling out: the 6 baseline runs that never found the credential were *not* misreports. The early exit kills children mid-flight, so hydra genuinely loses pairs it was about to test. The patch recovers those.

Single-target behaviour, all three scenarios:

| scenario | master | this branch |
|---|---|---|
| valid pair found | exit 0, no footer | exit 0, no footer |
| host does not resolve | exit 0, **no footer** | exit 255, footer |
| port closed | exit 255, no footer | exit 255, footer |

The middle row is a behaviour change beyond the bug report: master exits 0 on a scan whose only target never resolved. I read that as part of the same defect — the count that would have set `error` was the one being overwritten — but it does change an exit code, so flag it if you would rather keep that as-is.

### Notes

- No `clang-format` pass: the version I had available reformats unrelated pre-existing lines in `hydra.c` (the `SERVICE2`/`SERVICE`/`SERVICE3` macros, and three single-statement `if`s). It reported nothing on the changed lines, so the new code matches what your config produces.
- 3 hunks, 17 insertions, 8 deletions, `hydra.c` only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/hydra.c
+++ b/hydra.c
@@ -1507,26 +1507,37 @@ int32_t hydra_lookup_port(char *service) {
 
 // killit = 1 : kill(pid); fail = 1 : redo, fail = 2/3 : disable
 void hydra_kill_head(int32_t head_no, int32_t killit, int32_t fail) {
+  sigset_t chldmask, prevmask;
+  int32_t was_active;
+
   if (debug)
     printf("[DEBUG] head_no %d, kill %d, fail %d\n", head_no, killit, fail);
   if (head_no < 0)
     return;
-  if (hydra_heads[head_no]->active == HEAD_ACTIVE || (hydra_heads[head_no]->sp[0] > 2 && hydra_heads[head_no]->sp[1] > 2)) {
+  // This runs both from the main loop and from the SIGCHLD handler
+  // killed_childs(), which can interrupt the main loop anywhere in here. Block
+  // SIGCHLD so a head is torn down exactly once: otherwise both contexts see
+  // HEAD_ACTIVE, hydra_brains.active is decremented twice for the single
+  // increment in hydra_spawn_head(), and once it drops below the number of
+  // running heads hydra_check_for_exit_condition() ends the attack while
+  // children are still alive.
+  sigemptyset(&chldmask);
+  sigaddset(&chldmask, SIGCHLD);
+  sigprocmask(SIG_BLOCK, &chldmask, &prevmask);
+  // take the state once and use it for every decision below, so the counter
+  // update and the HEAD_ACTIVE -> HEAD_UNUSED transition cannot disagree
+  was_active = hydra_heads[head_no]->active == HEAD_ACTIVE;
+  if (was_active || (hydra_heads[head_no]->sp[0] > 2 && hydra_heads[head_no]->sp[1] > 2)) {
     close(hydra_heads[head_no]->sp[0]);
     close(hydra_heads[head_no]->sp[1]);
   }
   if (killit) {
     if (hydra_heads[head_no]->pid > 0)
       kill(hydra_heads[head_no]->pid, SIGTERM);
-    // only decrement for a head that is actually counted in hydra_brains.active.
-    // This function runs both from the main loop and from the SIGCHLD handler,
-    // so the same head can be killed twice; an unguarded decrement then drops
-    // the counter below the number of running heads and
-    // hydra_check_for_exit_condition() ends the attack while children are alive.
-    if (hydra_heads[head_no]->active == HEAD_ACTIVE)
+    if (was_active)
       hydra_brains.active--;
   }
-  if (hydra_heads[head_no]->active == HEAD_ACTIVE) {
+  if (was_active) {
     hydra_heads[head_no]->active = HEAD_UNUSED;
     hydra_targets[hydra_heads[head_no]->target_no]->use_count--;
   }
@@ -1554,6 +1565,7 @@ void hydra_kill_head(int32_t head_no, int32_t killit, int32_t fail) {
     //    NULL;
   }
   (void)waitpid(-1, NULL, WNOHANG);
+  sigprocmask(SIG_SETMASK, &prevmask, NULL);
 }
 
 void hydra_increase_fail_count(int32_t target_no, int32_t head_no) {
--- a/hydra.c
+++ b/hydra.c
@@ -1518,7 +1518,13 @@ void hydra_kill_head(int32_t head_no, int32_t killit, int32_t fail) {
   if (killit) {
     if (hydra_heads[head_no]->pid > 0)
       kill(hydra_heads[head_no]->pid, SIGTERM);
-    hydra_brains.active--;
+    // only decrement for a head that is actually counted in hydra_brains.active.
+    // This function runs both from the main loop and from the SIGCHLD handler,
+    // so the same head can be killed twice; an unguarded decrement then drops
+    // the counter below the number of running heads and
+    // hydra_check_for_exit_condition() ends the attack while children are alive.
+    if (hydra_heads[head_no]->active == HEAD_ACTIVE)
+      hydra_brains.active--;
   }
   if (hydra_heads[head_no]->active == HEAD_ACTIVE) {
     hydra_heads[head_no]->active = HEAD_UNUSED;
@@ -2339,7 +2345,7 @@ int main(int argc, char *argv[]) {
   size_t countinfile = 1, sizeinfile = 0;
   uint64_t math2;
   int32_t i = 0, j = 0, k, error = 0, modusage = 0, ignore_restore = 0, do_switch;
-  int32_t head_no = 0, target_no = 0, exit_condition = 0, readres;
+  int32_t head_no = 0, target_no = 0, exit_condition = 0, readres, active_heads = 0;
   time_t starttime, elapsed_status, elapsed_restore, status_print = 59, tmp_time;
   char *tmpptr, *tmpptr2, *tmpptr3;
   char rc, buf[MAXBUF];
@@ -4562,25 +4568,28 @@ int main(int argc, char *argv[]) {
   printf(" found\n");
 
   error += j;
-  k = 0;
+  // keep k as the target-derived count computed above: it is reported as a
+  // target count further down. The still-running worker heads are a separate
+  // quantity and get their own variable.
+  active_heads = 0;
   for (i = 0; i < hydra_options.max_use; i++)
     if (hydra_heads[i]->active == HEAD_ACTIVE)
-      k++;
+      active_heads++;
 
-  if (error == 0 && k == 0) {
+  if (error == 0 && active_heads == 0) {
     process_restore = 0;
     unlink(RESTOREFILE);
   } else {
     process_restore = 1;
-    if (hydra_options.cidr == 0 && k == 0) {
+    if (hydra_options.cidr == 0 && active_heads == 0) {
       printf("[INFO] Writing restore file because %d server scan%s could not "
              "be completed\n",
              j + error, j + error == 1 ? "" : "s");
       hydra_restore_write(1);
-    } else if (k > 0) {
+    } else if (active_heads > 0) {
       printf("[WARNING] Writing restore file because %d final worker threads "
              "did not complete until end.\n",
-             k);
+             active_heads);
       hydra_restore_write(1);
     }
     process_restore = 0;


gnuletik:
Good catch from the Copilot review — that was correct, and I've pushed a fix in 43661ce.

The `HEAD_ACTIVE` guard on its own was not idempotent. `SIGCHLD` could land between the check and the `active = HEAD_UNUSED` assignment — with a `kill()` syscall sitting in between, so the window was not even narrow — and both the main loop and `killed_childs()` would observe `HEAD_ACTIVE` and decrement for one increment.

Rather than reach for an atomic on `head_state_t`, I blocked `SIGCHLD` for the duration of the teardown and take the state once:

```c
  sigemptyset(&chldmask);
  sigaddset(&chldmask, SIGCHLD);
  sigprocmask(SIG_BLOCK, &chldmask, &prevmask);
  was_active = hydra_heads[head_no]->active == HEAD_ACTIVE;
  ...
  if (killit) {
    if (hydra_heads[head_no]->pid > 0)
      kill(hydra_heads[head_no]->pid, SIGTERM);
    if (was_active)
      hydra_brains.active--;
  }
  if (was_active) {
    hydra_heads[head_no]->active = HEAD_UNUSED;
    hydra_targets[hydra_heads[head_no]->target_no]->use_count--;
  }
  ...
  sigprocmask(SIG_SETMASK, &prevmask, NULL);
```

Blocking rather than a compare-and-swap because the same interleaving affects the other shared state in this function, not just the counter — the `sp[]` `close()` pair can double-close and `use_count` can double-decrement. One block covers all three. The function has a single early return (`head_no < 0`) and it is before any state is touched, so there is one exit path to restore the mask on. `was_active` then makes the pairing local and explicit, which is the invariant the review asked for.

Re-measured, same setup as the PR description — 16 concurrent processes, live `postgres:15.7`:

| | runs | credential found | bogus footer | non-zero exit |
|---|---|---|---|---|
| `4b552dc` | 320 | 320 | 3 | 3 |
| this branch | 640 | 640 | 0 | 0 |

Cumulative across all batches on this branch: 1760 runs, 0 bogus footers. Genuine failures still reported — unresolvable target prints the footer and exits 255, 10/10 under `--network none`.

No measurable cost from the two extra syscalls: 320 runs take 7s before and 8s after, which is inside the noise for this workload.

---

Separately, I have a deterministic regression test for this if you want it. It reaches `hydra_kill_head()` by renaming `main` away and including the source:

```c
#define main hydra_main_under_test
#include "hydra.c"
#undef main
```

and asserts the invariants directly — a second kill does not decrement again, a head that was never counted never decrements, the exit condition does not fire while a head is live, and the signal mask is restored on exit. It fails on `4b552dc` (`hydra_brains.active` reaches -1) and passes on this branch. No network, no concurrency, no timing.

It needs a `test:` target in `Makefile.am` mirroring the existing `hydra` rule (`$(OBJ)` is already the right object list), and it would be the project's first test, so I have deliberately kept it out of this PR — that is your call on whether you want test infrastructure at all. Say the word and I will add it here or as a separate PR. There is also a one-line black-box version needing no harness: `hydra -I -l x -p y postgres://nope.invalid` must print the footer and exit non-zero, which distinguishes the two builds 10/10.

vanhauser-thc:
thanks!

PR #1105, opened 2026-07-30
Replace VLA with heap allocation in hydra_tobase64() to prevent stack overflow

lxcxjxhx:
## Summary

- Fix VLA (Variable Length Array) stack overflow risk in `hydra_tobase64()`
- Replace stack-allocated VLA `unsigned char bof[i]` with heap-allocated buffer via `malloc(bufsize)`
- Prevent stack overflow crash when large `ufsize` values are passed
- Add proper `ree(bof)` on all return paths to avoid memory leaks

## Details

The original code used a C99 Variable Length Array (VLA) on the stack:

`c
unsigned char bof[i]; // i = bufsize, can be very large
`

When `ufsize` is large, this allocates a large buffer on the stack, which can cause a stack overflow and crash the program. This is a well-known security and stability risk in C programming.

### Fix

Replace the VLA with heap allocation:

`c
unsigned char *bof = malloc(bufsize);
`

All early return paths and the normal exit path now call `ree(bof)` to prevent memory leaks.


vanhauser-thc:
this is a false positive by your AI agent. bufsize is max 500 bytes, all good for the stack. unless you found a path where it would be > 8096 bytes (if so, please reopen)

PR #1106, opened 2026-08-03
Add payload pre-processing feature

smitverma:
# Feature : Add credential preprocessing for payload mutation

# Problem

Hydra currently inserts credential values into request templates without any transformation. This can make it difficult to test authentication flows that expect credentials in a derived representation, such as Base64-encoded, URL-encoded, normalized, or prefixed values.

Users have to generate alternate wordlists or create external wrapper scripts, which can be inconvenient for transformations that must occur only when constructing a request. External preprocessing also loses the distinction between the original credential candidate and the value placed on the wire.

This affects users testing services that use template placeholders such as `^USER^` and `^PASS^`, including HTTP form requests, RTSP URL-path authentication, and LDAP CRAM/DIGEST DN templates.

# Feature Overview

This PR adds a global `--preproc FIELD,TRANSFORM` option for applying bounded, built-in transformations to `USER` or `PASS` immediately before request-template placeholder expansion.

Supported transforms:

- `b64`
- `urlencode`
- `lower`
- `upper`
- `prefix:VALUE`
- `suffix:VALUE`

Rules may be chained within one option and repeated across options. For example:

```bash
hydra --preproc PASS,b64,urlencode ...
```

This applies Base64 encoding first, then URL encoding on the password payloads.

The feature affects only template substitution. Hydra continues to retain, schedule, deduplicate, report, and restore the original credential candidates rather than transformed values.

# Implementation Details

### `hydra.c`

- Added `--preproc FIELD,TRANSFORM` to help output and command-line handling.
- Added a portable long-option normalization step that converts:
  - `--preproc PASS,b64`
  - `--preproc=PASS,b64`
  
  into an internal `-Z` option before Hydra’s existing `getopt()` loop processes arguments.

- Added validation for:
  - Invalid fields; only `USER` and `PASS` are accepted.
  - Empty transformations.
  - Unsupported transformations.
  - Excessive rule or transform counts.
  - Prefix/suffix values exceeding the configured limit.

- Rules are stored in a dedicated global `hydra_preproc_config`, rather than extending `hydra_option`. This avoids changing the existing serialized `hydra_option` layout, which contains pointer fields.

- Added a marked preprocessing block to restore files:
  - Newly written restore files persist preprocessing rules.
  - The updated binary accepts older restore files that do not include the marker and defaults to no preprocessing rules.
  - Restore data is validated before use.

### `hydra.h` and `hydra-mod.h`

Added fixed-size configuration types and shared helper declarations:

- `hydra_preproc_field_t`
- `hydra_preproc_transform_type_t`
- `hydra_preproc_transform`
- `hydra_preproc_rule`
- `hydra_preproc_config`
- `hydra_preprocess_credential()`
- `hydra_replace_credential_placeholders()`
- `hydra_preproc_uses_transform()`

The fixed limits are:

- Up to 32 rules.
- Up to 16 transforms per rule.
- Prefix/suffix values up to 255 characters plus a terminator.
- Intermediate and expanded values capped at 6000 characters.

### `hydra-mod.c`

Added centralized request-time transformation support:

- `hydra_preprocess_credential()` applies matching rules in command-line order.
- `hydra_replace_credential_placeholders()` expands all `^USER^` and `^PASS^` occurrences using transformed values.
- Internal helpers safely:
  - Replace all placeholder occurrences.
  - Base64 encode.
  - Percent-encode URL values using RFC 3986 unreserved characters.
  - Convert case.
  - Add prefix or suffix data.
  - Bound output sizes and return failure instead of overflowing buffers.

The helpers allocate owned temporary strings and callers free them after request construction.

### `hydra-http-form.c`

HTTP form request construction now preprocesses both credentials before generating:

- `^USER^`
- `^PASS^`
- `^USER64^`
- `^PASS64^`
- Matching user-defined header placeholders

`^USER64^` and `^PASS64^` are Base64 representations of the preprocessed credential values.

Existing form encoding remains in effect for normal credential placeholders. If the final transform for a field is `urlencode`, that transformed value is inserted directly to avoid an additional form-encoding pass.

The module rejects oversized transformed values and skips the candidate rather than constructing an unsafe request.

### `hydra-rtsp.c`

RTSP URL-path substitution now uses `hydra_replace_credential_placeholders()` instead of its local one-occurrence replacement logic. This applies preprocessing to RTSP paths containing `^USER^` or `^PASS^`.

RTSP Basic and Digest authentication remain unchanged because they do not use credential placeholders.

### `hydra-ldap.c`

LDAP CRAM-MD5 and DIGEST-MD5 DN template expansion now uses a temporary processed template rather than overwriting `miscptr`.

This preserves the original module option template across attempts and applies preprocessing only to placeholder-derived DN values.

### `hydra.1`

Updated the synopsis and option documentation with:

- Option syntax.
- Supported fields and transformations.
- Chaining behavior.
- The request-time-only behavior.
- HTTP form URL-encoding behavior.

# Usage Examples

`--preproc` is a global Hydra option and must be provided before the target/service positional arguments.

```bash
hydra --preproc PASS,b64 -l admin -P passwords.txt target http-post-form \
  "/login:user=^USER^&pass=^PASS^:F=Invalid"
```

Base64-encodes each password immediately before it replaces `^PASS^`.

```bash
hydra --preproc USER,b64 --preproc PASS,b64 \
  -l test -p test target http-post-form \
  "/login:username=^USER^&password=^PASS^:F=Invalid"
```

Base64-encodes both credentials independently before request construction.

```bash
hydra --preproc PASS,b64,urlencode \
  -l testuser -p testpass target http-post-form \
  "/:username=^USER^&password=^PASS^:F=Invalid"
```

Chains transformations. `testpass` becomes `dGVzdHBhc3M=` and is URL encoded for insertion into the HTTP form body.

```bash
hydra --preproc PASS,lower \
  -l testuser -p TeStPass target http-post-form \
  "/:username=^USER^&password=^PASS^:F=Invalid"
```

Normalizes the password to `testpass` before replacing `^PASS^`.

```bash
hydra --preproc 'PASS,prefix:PRE-,suffix:!' \
  -l testuser -p testpass target http-post-form \
  "/:username=^USER^&password=^PASS^:F=Invalid"
```

Chains a prefix and suffix so the request contains `PRE-testpass!`.

```bash
hydra --preproc USER,urlencode \
  -l 'user@example.com' -p password target rtsp \
  -m "/stream?user=^USER^&pass=^PASS^"
```

URL-encodes the username when expanding RTSP URL-path placeholders.

# Testing

No automated test suite or test target exists in the current project build configuration, and this PR does not add one.

The following checks were performed:

```bash
./configure
make -j2
```

Configured and built the project successfully.

```bash
git diff --check
```

Verified the final diff has no whitespace errors.

CLI parser validation was manually checked:

```bash
./hydra --preproc FOO,b64 -h
```

Confirmed invalid fields are rejected.

```bash
./hydra --preproc PASS,rot13 -h
```

Confirmed unsupported transforms are rejected.

```bash
./hydra --preproc=USER,lower,prefix:DOMAIN \
  --preproc PASS,b64,urlencode -h
```

Confirmed repeated rules, chaining, and both supported long-option spellings are accepted.

A temporary localhost HTTP form fixture recorded raw POST bodies and parsed fields for these five request-time cases:

1. `PASS,b64`
   - Parsed password: `dGVzdHBhc3M=`

2. `USER,b64`
   - Parsed username: `dGVzdHVzZXI=`

3. `PASS,lower` with `TeStPass`
   - Parsed password: `testpass`

4. `PASS,prefix:PRE-,suffix:!`
   - Parsed password: `PRE-testpass!`

5. `PASS,b64,urlencode`
   - Raw form body contained `dGVzdHBhc3M%3D`.
   - Parsed password was `dGVzdHBhc3M=`.

The localhost fixture and generated request log were removed after verification.

# Backward Compatibility

Commands that do not use `--preproc` retain existing behavior:

- Credential wordlists are not modified.
- The raw credential scheduler and worker-pipe format are unchanged.
- `-C`, `-e`, `-u`, brute-force generation, retry behavior, duplicate detection, reporting, and output formatting continue to use raw credentials.
- Non-template protocol authentication paths remain unchanged.
- Existing template substitutions behave as before when no preprocessing rules are configured.

This adds one new global option and does not require configuration changes for existing users.

The updated binary can read pre-existing restore files without preprocessing data. Restore files written by the updated binary contain an additional marked preprocessing block and should be resumed with a binary that includes this feature.

# Security / Performance Considerations

- The feature accepts only a fixed allowlist of transformations. It does not execute scripts, shell commands, expressions, or arbitrary user-provided code.
- Transformation values and generated output are bounded to avoid uncontrolled request expansion.
- Invalid or oversized transformed values cause the attempt to be skipped rather than being inserted into request buffers.
- Base64 encoding is an encoding operation, not encryption or credential protection.
- Preprocessing occurs per request attempt and introduces temporary allocation/copying overhead only for template-based paths using this feature.
- Existing no-`--preproc` paths do not invoke the new transformation flow.

# Review Notes

- The implementation intentionally applies preprocessing only at existing `^USER^` and `^PASS^` expansion points. It does not alter protocol-native Basic, Digest, SASL, or other non-template credential handling.
- HTTP form handling treats a final `urlencode` transform specially to prevent double encoding. Maintainers may want to confirm this interpretation remains the desired contract for chains that combine URL encoding with other transforms.
- CLI options follow Hydra’s existing positional convention: `--preproc` must appear before the target/service arguments.
- My future update plans include adding support for encoding and hashing pre-processing - eg. ROT13, MD5, SHA1, etc.


smitverma:
Hi @vanhauser-thc, hope you're doing good. Is there any update on this PR? 

I personally found the feature very helpful in many payments and geography related APIs that expect data in various encodings.

If there's any changes you recommend, I can do them!

Thank you.

vanhauser-thc:
Hi!
thanks for your PR!
I am torn on it honestly. it brings complexity into the user experience and code, and then it is not possible to be 100% ("for me it needs to be uppercase hex encoded!"). also I am looking forward to the day that 1-2 really good competitors are out there so I can finally archive this code.
I will need to think more about this (and activated copilot to take a look)

smitverma:
@vanhauser-thc 

Hello,
I understand your concerns and am thankful for your time to review this PR.
Looking forward to your decision!

Also what in this world are those update-readme PRs 😭

